PR #23, opened 2018-01-25
Type stuff

ajs6f:
Two commits here: the first will be uncontroversial, I think-- it just adds in some type params to avoid compiler warnings. Maybe there's a reason not to, and if so, let's just skip it.

The other widens the return type of `ResourceService::get` (in both forms) in order to deal with Java's craptacular generics and their lack of proper co/contravariance control. Because `ResourceService::get` returns `Optional<Resource>`, I can't impl it with `Optional<MyTypeThatImplsResource>` because Java doesn't understand that `Optional<MyTypeThatImplsResource> "extends" Optional<Resource>`. So now I can, at the cost of making some test mocking code look pretty gross... this lets impls of `ResourceService` use `get` in their own code and get the actual type of `Resource` at stake. _This_ is useful because `MyTypeThatImplsResource` can have extra fresh members (in my case for Cassandra, a member that gives the Basic Containment container for that resource).

Does that make sense?

### DIFF
--- a/trellis-api/src/main/java/org/trellisldp/api/ResourceService.java
+++ b/trellis-api/src/main/java/org/trellisldp/api/ResourceService.java
@@ -13,7 +13,6 @@
  */
 package org.trellisldp.api;
 
-import static java.util.Optional.empty;
 import static java.util.Optional.of;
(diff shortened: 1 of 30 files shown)


acoburn:
This all looks good to me.

PR #30, opened 2018-02-01
Rearrange the furniture

ajs6f:
Introducing new types for persistence, but not yet weaving them all the way through the module stack.


acoburn:
This looks great. I'd like to add some tests for the DatasetCollector, but that can be done separately.

PR #31, opened 2018-02-01
Factoring new persistence types through `-http`

ajs6f:
DO NOT MERGE WITHOUT CONVERSATION

This is my preliminary sketch as to how to push the new persistence types up through`-http`. I'm a little unsure about (at least) one thing: I'm using the HTTP IRI to store audit data. Is that right or would we rather use the internal non-HTTP IRI?


acoburn:
At present, the `AuditService` object is injected via the ServiceLoader, which typically will pick up the `DefaultAuditService` class. But that class doesn't know how to act on `::add` (it always returns false). Here, I suspect you'll need to add a (nullable) AuditService object in the LdpResource constructor, since the ServiceLoader likely won't be sufficient.

ajs6f:
Returning `false` all the time is okay for the mo-- I actually don't check, because I wanted to get the calculations of what to store right first. But I see what you mean. When I add a commit to this (tomorrow) to actually check (and fail the request if audit fails to store) I will need to provide a test-usable impl. 

Would you be weirded out if I just subclassed `DefaultAuditService` into a `TestAuditService` that just returns `true` for `::add`? Just for testing, of course.

acoburn:
I'd advocate for moving the AuditService to the constructor. Then you can mock that method to your heart's delight.

ajs6f:
Okay, but is that with the idea of allowing a `null` for that param to indicate "no audit"?

acoburn:
On thinking of this (esp. in the context of CDI) I wonder if it might not be better to have a `NullAuditService` implementation -- one that generates no quads and does nothing with `::add`. That is, instead of wrapping the `AuditService` in an `Optional`.

ajs6f:
Yes. A "null object" makes sense here, especially because the behavior is so simple. Commit forthcoming.

ajs6f:
I don't like putting a test package into `-audit` like that. But I don't much like one module relying on another's test artifact, either, and I don't want to duplicate that `TestAuditService`. Urg.

ajs6f:
Ok, I feel a little happier about this. Now we have a complete separation between pathways for "immutable" (currently just audit) and "mutable" (user-ish, LDP-ish) data unless and until the integrator combines those paths using a concrete type that impls both `AuditService` and `ResourceService` (and thereby indirectly `AppendService` and `ReplaceService`). 

And that was my whole point. Now we can clean this up and make it right and maybe it'll get ready for merging.

I think the first thing is moar tests. Time for me to start reading a _lot_ of test code. :)

ajs6f:
Just added some more tests, but also changed up the ctors for `TriplestoreResourceService`. Frankly, I don't like what I did, adding a `boolean` param. But there has to be some way for the integrator to distinguish between three cases:
1. Here is an `AuditService` I want to use.
2. I want to use this very `TriplestoreResourceService` that I am creating as the `AuditService`.
3. Go to the `ServiceLoader` for the `AuditService`.

Yet again, the urgency of DI increases.

ajs6f:
Well, that's all the work I'll be doing for the mo. Travis is failing with
```
org.trellisldp.app.TrellisApplicationTest > TrellisApplicationTest.testPostBasicContainer() FAILED
    org.opentest4j.AssertionFailedError
```

which is bizarre because there is no such method in that class that I can see (although it certainly _sounds_ like something that would go in that class). I'll wait for @acoburn to comment.

acoburn:
The `TrellisApplicationTest::testPostBasicContainer` is from a commit I just added: https://github.com/trellis-ldp/trellis/commit/426cba3d00708ec9d391f06c2663c5dca35c9e2d

You might need to rebase from master to pick up that test and/or do a `./gradlew clean check`

ajs6f:
Ah, cool. Rebasing.

ajs6f:
Weird. It's failing because the ETag for a resource is not changing even though the test POSTs several times to it to create new resources.

acoburn:
What a pain. Those same tests passed an hour ago: https://travis-ci.org/trellis-ldp/trellis/builds/336688987

I have found that there is some level of flakiness in the time-based tests on CI infrastructure. I have been really reluctant to add `sleep(...)` to force certain gaps in the timing. One of the issues I've noticed (and this tends to happen more on JDK8 Windows environments) is that `Instant` not only has only millisecond precision but it doesn't always seem to proceed linearly (!).

If these tests start to get really flaky, then we should strengthen the tests, but for now, if it builds locally and only fails intermittently on CI infrastructure, I wouldn't worry too much about it.

ajs6f:
It fails reliably for me, but I've been staring at this stuff since my last comment and I can't figure out the prob. I guess the interactions are happening fast enough that the last-modified date isn't changing, since that and the identifier are what builds the ETag.

I agree that `Thread::sleep` is really weak cheese. But if the test really does depend on time (i.e. ETags are made of nothing but time and identity, no resource-content involved) then I'm not sure of a better alternative.

acoburn:
OK, I'll add some sleep(...) code. Incidentally, are you building with JDK 8 or 9?

ajs6f:
8. You think it's relevant? I haven't yet built up a reliable 9 tooling on this box, but I certainly can. I need to do that anyway for Jena.

acoburn:
I think it's relevant b/c a JDK 8 Instant has millisecond precision, but a JDK 9 Instant has nanosecond precision. And since ETags are generated from `Instant` values (plus the resource IRI), if the change occurs w/i the same millisecond, then the ETag won't change. In a _real_ scenario where network traffic is involved, this really wouldn't ever be an issue, but for entirely in-memory, local network tests, that issue can definitely emerge.

ajs6f:
Makes sense.

Ok, I'll go detour to build up a Java 9 plant (which, like I say, I need to do anyway).

acoburn:
This _should_ do the trick: https://github.com/trellis-ldp/trellis/commit/d78ccd34c35ed5d3716c3aba8d0ac66eefac3283

ajs6f:
Thanks, building with J9 (just for grins), then will rebase and keep moving.

ajs6f:
P.S. Any opinion about the ctors for `TriplestoreResourceService`? Or just let it ride until we wrestle with wiring?

acoburn:
Re: ctors, I'd prefer to keep boolean values out of the `TriplestoreResourceService`.

ajs6f:
Agreed, but how _do_ we distinguish between the three cases... an `enum`? A factory (gross!)?

acoburn:
I would distinguish b/t the three cases by allowing only one of them:

  1. Here is an AuditService I want to use.
  2. ~~I want to use this very TriplestoreResourceService that I am creating as the AuditService.~~
  3. ~~Go to the ServiceLoader for the AuditService.~~

This way, there is a single constructor for http.LdpResource. It may mean that the same object is passed into the ctr twice, but (if I am not mistaken) CDI will handle that just fine.

That is, in the application code:

```java
final TriplestoreResourceService svc = new TriplestoreResourceService(...);
...
environment.jersey().register(
    new LdpResource(svc, ioService, binaryService, svc, baseUrl));
```

(we need to do something better w/ `baseUrl`, but that's a separate issue.)

ajs6f:
Nope, I would have done that, but `TriplestoreRS` _requires_ an `AuditStore` even in its own construction. (In `init()`.) That's the real problem. (And yes, with CDI or other injection, the problem goes away because we can make `init()` a `@PostConstruct`). That's why I'm wondering if we let it ride.

Or I can get fancy and factor `init()` around to be lazier... but that gets weird quickly. You get into threadsafety and other things that just aren't what we're trying to do anyway...

acoburn:
Oh right. If _that_ is the issue, I'd just refactor the init method to remove those audit quads. I think they are of questionable value anyhow.

ajs6f:
Oh, for cryin' out loud. If'n I'd known that, I would have just...

This is why distributed teams are hard.

Okay, I may not fix the annoying ETag thing (which still seems to be failing) but I will send another commit cleaning up the `TriplestoreRS` craziness.

acoburn:
The last-modified issue should now be resolved; at least Appveyor is happier. But if those latest commits in master don’t resolve things for your local builds, please let me know.

acoburn:
And BTW, I believe your last commit is failing due to a javadoc error (an unused parameter tag)

ajs6f:
Yep, that would do it. Let's see if I get a build now.

ajs6f:
Nope, the same ETag failure. Hm... you did put `sleep`s in there, so now it's really a mystery.

acoburn:
OK, I'm going to fix this correctly with awaitility and a better ETag generation function.

ajs6f:
I have successfully made work for both of us on the weekend. Go me!

acoburn:
See if https://github.com/trellis-ldp/trellis/commit/8e3c733b5d80936433398c59672a130508805ff2 helps at all.

acoburn:
It appears that I was addressing an issue that was different than the one you were encountering.

I believe the issue that's behind the error here is the fact that the Triplestore impl relies on audit quads to determine if a given `ResourceService::put` command is a Create, Update or Delete action. Given that (with this refactor), Audit quads won't be part of the Dataset passed into the method, the code will not ever pass through the `isCreate` or `isDelete` blocks, which means that parent containers won't ever be updated, which of course means that ETags won't change when containment triples change.

Here's a suggestion: perhaps the `ResourceService::put` method needs an additional argument for the Activity type. The method can accept any IRI, but the HTTP layer would know enough to set `AS.Update`, `AS.Delete` or `AS.Create`, as appropriate. Then, the code in the TrellisResourceService won't have to look through the Dataset to find the appropriate "audit" quads indicating what the specific operation was.

That would make the new method:

    Future<Boolean> put(IRI identifier, IRI ixnModel, IRI activity, Dataset dataset);

And, possibly, the `ReplaceService<T>` might become:

    Future<Boolean> put(IRI activity, T resource);

Alternately, it would be possible to add a Triple to the Dataset (e.g. in the default graph) which indicates the activity type, but I think it seems like a clearer contract if a dedicated parameter is used. Also, if you prefer to define an `enum` for this in the `trellis-api` (rather than using terms from the ActivityStream vocabulary), that's also fine.

ajs6f:
Ah, okay, cool. It would have taken me a day or three to figure that out, so thanks!

Let me ruminate on this a bit and offer a commit in another day or so. Is that cool?

acoburn:
Take your time. Thanks for working on this.

ajs6f:
Oh, it's all selfish. I just want to make the Cassandra impl easier for me. :wink:

acoburn:
@ajs6f I added a commit to master that introduces a conflict to your PR. It should be trivial to resolve, though. In your PR, the conflicting line (in TrellisApplication.java) should now be:

```java
final TriplestoreResourceService resourceService = new TriplestoreResourceService(rdfConnection, idService,
                mementoService, notificationService);
```

ajs6f:
Which module should be responsible for updating the parent? Should it indeed be the persistence (aka the `ResourceService` impl) or the HTTP layer? Especially since the parent modification may be occurring asynch to the child modification, should the HTTP layer issue two messages into the persistence service? Then we stay with current `::put` but call it twice, if you see what I mean.

acoburn:
I'm inclined to keep that logic in the `ResourceService` just so that impl has more flexibility on how it deals with parent and membership resources. A big part of my rationale here has to do with making it easy for async processing but also because I have no idea what a reasonable interface would even look like for updating indirect containers (it certainly wouldn't be "PUT").

ajs6f:
Aren't indirect containers just `Resource`s? Why wouldn't "PUT" work? Yes, you'd have to retrieve first, but I don't see that as a bad thing...

acoburn:
PUT is problematic because you would need to fetch all child resources of the indirect container, read the triples for each of those child resources, finding the relevant `ldp:insertedContentRelation` property, accumulate those into a new dataset and then `PUT` into the membership resource. And you have to hope that no other child resource is modified in the meantime. It is problematic in terms of both concurrency (do you have to lock in the HTTP layer?) and memory usage (what about an indirect container with _a lot_ of child resource?). And that process needs to happen every time any child resource is touched. Direct containers are not nearly so challenging, but for large containers, you'd still need to collect a lot of resources into a single dataset while not allowing concurrent modifications.

ajs6f:
I'm lost-- why would you need to touch _any_ of the child resources (other than the on you started with, of course)? Are you starting from the assumption that the child resources (for an indirect container) keep their own containment triples with them? (as per our 'phone discussion earlier today)

ajs6f:
I can't find any way to read the [Indirect Container section](https://www.w3.org/TR/ldp/#h-ldpic) so that triples are created in children at any time (except by the usual method of putting them in an RDF representation). The only triples that must be created that I can find are under "5.5.2 HTTP POST" and are in the container.

As I understand the matter (which is badly at best) Indirect Containers only offer the possibility of "aliasing" the container-contained relationship. For example, [this interaction](https://www.w3.org/TR/ldp-primer/#IndirectCreate) shows no triples in the contained resource that weren't put there explicitly by the client, just new triples in the container.

acoburn:
If you start with the assumption that you are modifying an arbitrary resource (create, delete _or_ update), you first need to check the parent container to see if the parent is an indirect container. If it _is_ an indirect container, then you'll need to find out where the membership resource is located (defined in the parent container via ldp:membershipResource). If you're using the semantics of `put`, then you'd need to update a subset of the membership resources for that container (a resource may be the target of multiple Direct/Indirect containers). So how do you know which membership triples to update? If the persistence layer generates the membership triples dynamically, then the only property to update is `lastModified`, since the resource representation will be different. However, if the persistence layer doesn't generate the membership triples dynamically (i.e. it stores them in advance), then you'd need to figure out what all of the membership triples would be by reading _all_ of the child resources. That would be really computationally expensive for containers with a lot of resources.

And depending on the persistence layer implementation, making the update may be very, very simple -- it just isn't a PUT. In the triplestore and the kafka-based persistence impls, the membership triples are handled very differently, but both involve mechanisms that are most certainly not `PUT`.

The case of direct containers is easier, but again, by putting the logic in the HTTP layer, you force a certain model of persistence w/r/t membership triples that IMO is much more complicated than it needs to be.

acoburn:
I should point out that it is also possible that I am completely misunderstanding indirect containment.

ajs6f:
You and me both. 

Regardless, the fact that multiple containers might be involved is telling-- that would require some potentially-intractable work. Now it worries me that `ResourceService::getContainer` returns a single IRI. But I think you wrote this backwards:

>  However, if the persistence layer doesn't generate the membership triples dynamically (i.e. it stores them in advance), then you'd need to figure out what all of the membership triples would be by reading all of the child resources.

It's the _dynamic_ case that is (potentially) expensive. The static case is easy. You don't ever have to go the child resources because all of the containment info is either stored with the parent (static) or in some other way. But I can see your point about wanting to be agnostic to how. We just don't know how people might store stuff.

Let me think about this some more. I'm increasingly convinced that making `ResourceService` do "create/replace/delete" as a single primitive `::put` is not going to work the best in the long run, but I haven't got the right alternative. Maybe just three methods: `::create`,`::replace`, and `::delete`?

acoburn:
`::create`, `::replace` and `::delete` makes a lot of sense to me.

ajs6f:
Gordian knot, meet Alexandrian sword. I'll make a commit up over the next day-ish so that we can look at something concrete.

ajs6f:
For `::delete`, do you know off the top of your head where I could look to see what the upper layers expect to get back from a resource after it has been deleted? I.e. what quads...

ajs6f:
Hang on, stupid question. Just `::delete(IRI id, Dataset quads)` and let the upper layers decide. Never mind, all good.

ajs6f:
There has been some discussion here and there about whether it's reasonable to let clients change the interaction model of an extant resource. Can I assume that Trellis will take its usual agnostic position on that?

acoburn:
via `HTTP PUT`, you can change a resource to an LDP subclass of the existing resource.

https://github.com/trellis-ldp/trellis/blob/master/trellis-http/src/main/java/org/trellisldp/http/impl/PutHandler.java#L174-L178

(A client would use link headers to do this). I'm not wedded to that feature, but it seems like a decent middle ground between allowing everything (binaries can become containers?) and allowing nothing (no changes ever).

Of course, a client can always `DELETE` a resource and then recreate it as something else.

ajs6f:
Okay, cool. (The practical question for me is whether there is an `IRI ixnModel` param for `::replace`, and the answer is yes, there is.)

ajs6f:
Tests are failing, but this shows where I'm going.

trellis-http/src/main/java/org/trellisldp/http/impl/PutHandler.java
```diff
@@ -236,8 +237,11 @@ public ResponseBuilder setResource(final Resource res) {
                         .forEachOrdered(dataset::add);
                 }
             });
-
-            if (resourceService.put(internalId, ldpType, dataset.asDataset()).get()) {
+            // TODO is this the best we can do? what concurrency errors are lurking?
```
acoburn: it is possible for two nodes to have simultaneous `::create` operations. It is therefore possible for the persistence layer to duplicate containment triples on the parent, but it is also possible for that layer to deal intelligently with that condition. E.g. the kafka-based impl uses ZK as a coordinating agent to prevent against that condition. And for synchronous implementations, the tricky parts are not so hard to avoid (at least they weren't for the triplestore impl).

ajs6f: Yeah, the question for me is how much the upper layers (above `ResourceService`) want to do about that. My inclination is: not very much at all!

acoburn: Yes, I completely agree. I don't want to see any locking/coordinating code anywhere in the `-api` or `-http` layer.

trellis-http/src/main/java/org/trellisldp/http/impl/PutHandler.java
```diff
@@ -239,8 +237,24 @@ public ResponseBuilder setResource(final Resource res) {
                         .forEachOrdered(dataset::add);
                 }
             });
+            // TODO is this the best we can do? what concurrency errors are lurking?
+            Future<Boolean> success = resourceService.get(internalId).isPresent()
```
acoburn: Actually, this check here is overkill. A better formulation would be:

```java
Future<Boolean> success = nonNull(res)
                  ? resourceService.replace(internalId, ldpType, dataset.asDataset())
                  : resourceService.create(internalId, ldpType, dataset.asDataset());
```

Basically, this very same check is already performed at a higher level in `LdpResource`, and depending on the result it either calls `::setResource` with the retrieved `Resource` or it calls the no-argument `::setResource` (which is equivalent to calling `setResource(null)`).

If we want to be more aggressive about avoiding conflicting writes, we could require an `If-Match` header for all PUT-for-update requests. (LDP describes this at the level of `should` -- Trellis supports `If-Match` but doesn't require it for updates).

ajs6f: That's good to hear (that the check can be so much less expensive). I'm not a big fan right now of adding restriction to LDP-- I'd rather leave collision avoidance to the backend for now unless we find out that everyone does it exactly the same way. (Which could very well happen...) Keep HTTP as stateless and unconversational as possible!

ajs6f:
Still failing `trellis-app` tests, but I think the persistence types make more sense. The tests should start passing if I can make sure that audit info is getting recorded and passed around correctly.

ajs6f:
P.S. This PR is getting crazy messy, so once it seems good, I may rewrite it cleanly. But first it has to get good.

acoburn:
OK, I'm refactoring those `trellis-app` tests to make use of JUnit5 `@Nesting`, which helps keep the code considerably better organized. Just an FYI to not touch the application test code until after I commit these changes (otherwise there will be mad rebasing).

acoburn:
Just a heads-up: I found a bug in the triplestore layer (it affects direct/indirect containers whose member resource is a hashIRI). I have a fix for it, but it will result in some rebasing on your part.

acoburn:
@ajs6f ping
just wanted to check where this PR stands

ajs6f:
ARRRRGGG. Tied up with too many operational things.

Two tasks remain:
1. make tests pass and
2. offer a `ResourceService` impl that accepts a `MutableDataService<IRI, Resource>` and a `ImmutableDataService<IRI, Resource>` (presumably as injectable cstor params) and just drops all calls through to them. That would be how someone who wants to use genuinely immutable persistence for audit would do it.

The second task should be trivial. The first I don't _think_ should be too bad. The failing tests are all in `TripleStoreResourceService` (and the app tests, but they are failing on the same accounts) and have to do with tests using audit triples. I don't think it should be hard to make them pass, but I don't know when I will have a chance to look hard at it. Certainly not until later this coming week. 

It wouldn't bother me to put this on ice for a while, if you want to clear the deck to simplify things going forward for the next few weeks. There are some good ideas in it that are unrelated to the main thrust of the PR but they will resurface ifwhen they again become visibly useful. Or you can let it hang (potentially until the week of the 26th) and I will get back on it.

acoburn:
No worries. I have rebased it with the latest commits from master. There were also some failing tests in `trellis-http`, which I've fixed in the last commit. As time allows, I can work on fixing the -triplestore and -app tests, if you'd like to work on the second task. That will also give me an opportunity to become more familiar with the changes.

ajs6f:
That sounds great! You should def do that careful review, because alongside rthe decent ideas, there is a lot of thrash aftermath in here. I will sort some scheduling stuff out over the weekend and by Monday I should have a really good idea of when I will get the second task done.

ajs6f:
Oh, &*^&*%**^. Can't believe I didn't catch `::create`=>`::replace`.

ajs6f:
So, a first draft of a `ResourceService` that passes through to underlying (im)mutable persistence. The trickiness is, of course, in the fact that the underlying services have extremely simple and intentionally generic contracts written around `Resource`s, whereas (ironically) `ResourceService` itself deals in `Dataset`s and `IRI`s. So I introduced some private classes to contain datasets and IRIs. The one that merges data seems unobjectionable to me, but the one that packages data as `Resource`s does have state. Given that the state exists only within the bounds of a single persistence callstack, it doesn't worry me too much, but a stupid `(Im)MutableDataservice` _could_ theoretically_ let that state leak out somewhere...

What do you think? Should I write tests (with an example)? I _think_ we have now obtained the original goal of this PR, which was to allow completely independent paths through from HTTP to persistence for immutable and mutable data. Now, for the Cassandra impl, I can impl the two `(Im)MutableDataservice` types, with `MutableDataservice` just replacing quads in Cassandra and `ImmutableDataservice` just appending quads, very simple on each side, and a `ResourceService < JoiningResourceService` that just layers in the fact that I'm not going to imply Memento, at least not now. Or someone could put mutable stuff in Cassandra and immutable stuff in Kafka, or even stick in filters and the like for one side without messing with the other.

P.S. the name `JoiningResourceService` is crap. It should be better.

P.P.S. If/when this gets merged, I'm going to squash and cherrypick around a bit to clean up the history so that people don't have to wade through me adding formatting changes and license headers and so forth.

trellis-api/src/main/java/org/trellisldp/api/ResourceService.java
```diff
@@ -56,7 +66,37 @@
...
+
+    /**
+     * Delete a resource from the server.
+     *
+     * @param identifier the identifier for the new resource
+     * @param ixnModel the new LDP interaction model for this resource
+     * @param dataset the dataset
+     * @return whether the resource was deleted
+     */
+    Future<Boolean> delete(IRI identifier, IRI ixnModel, Dataset dataset);
```
acoburn: I have been wondering if this method shouldn't be just `delete(IRI)`. What do the `ixnModel` and `dataset` mean in this context? This would clearly have implications for other interfaces.

trellis-api/src/main/java/org/trellisldp/api/JoiningResourceService.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Uses two underlying persistence services (an {@link ImmutableDataService} and
+ * a {@link MutableDataService}) to provide the {@link ResourceService} API.
+ *
+ * @author ajs6f
+ * @see ImmutableDataService
+ * @see MutableDataService
+ */
+public abstract class JoiningResourceService implements ResourceService {
```
acoburn: Would `AbstractResourceService` be a better name? (or `AbstractResourcePersistenceService` though that's getting pretty long). "Joining" does communicate the connection between mutable and immutable data. There's also no harm in keeping the name as it is and modifying it later when someone finds a better name.

trellis-audit/src/main/java/org/trellisldp/audit/DefaultAuditService.java
```diff
@@ -42,13 +38,9 @@
...
  * graph.
  * 
- * @implSpec This implementation of AuditService does not persist audit
- *           information. Subclass and override {@link #add(Resource)} to add
- *           persistence.
- * 
  * @author acoburn
  */
-public class DefaultAuditService implements AuditService {
+public abstract class DefaultAuditService implements AuditService {
```
acoburn: Why is this abstract? Hasn't the `::add` method been moved elsewhere?

ajs6f: Yup, looks like an oversight which I will now fix.

trellis-api/src/main/java/org/trellisldp/api/JoiningResourceService.java
ajs6f: I like `AbstractResourceService` better by a bit. But maybe it _is_ better to leave it be for now. If it doesn't really bug you I will leave it be.

trellis-api/src/main/java/org/trellisldp/api/ResourceService.java
ajs6f: The problem is that deleting something doesn't. It tombstones it, and we need to know with what to build the tombstone. Unless you are saying that from the HTTP POV, a `DELETE` message should use `::replace` on the backend? (Which actually makes a lot of sense.)

trellis-api/src/main/java/org/trellisldp/api/JoiningResourceService.java
acoburn: Feel free to leave this as-is.

ajs6f: k, then I'll move on to some tests/examples.

trellis-api/src/main/java/org/trellisldp/api/ResourceService.java
acoburn: We would still want to distinguish between an empty `HTTP PUT` (one that removes all triples) and a `HTTP DELETE` (one that removes triples and sets a tombstone -- though, a persistence impl could choose to really delete the resource, sans tombstone).

At present, the Triplestore impl just ignores the `ixnModel` and `dataset` values, but that also doesn't mean that's what every impl needs to do. I'm fine with leaving this as-is because an implementation still has a lot of control over these methods.

acoburn:
So, I'm 👍  on the state of this PR. I will leave the merging to you so that you can squash/cherry-pick, etc, as you see fit.

trellis-api/src/main/java/org/trellisldp/api/ResourceService.java
ajs6f: Yeah, the thing is to both distinguish between `PUT` and `DELETE` and also to distinguish between `::replace` and `::delete`. Can we say for sure that any HTTP `PUT` will result in a `::replace` and any HTTP `DELETE` will result in a `::delete`? If so, we need to leave these params in. If both `PUT` and `DELETE` will map to `::replace` (with `DELETE` just becoming a special case of "replace with a tombstone") then we can probably drop the params.

acoburn: I am actually noticing a condition that speaks exactly to your point: https://github.com/trellis-ldp/trellis/blob/FinishNewPersistence/trellis-http/src/main/java/org/trellisldp/http/impl/DeleteHandler.java#L92-L101

When an ACL graph is deleted (via DELETE), the user-managed graph remains intact. Furthermore, the `baseURL` value needs to be sent along for use with the event stream. (This also means that there is a bug in the triplestore impl). Either way, the signature definitely needs to stay the way it is.

ajs6f: Okay, that works for me! On to some tests and some Git-fu.

trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -141,58 +140,71 @@ private static RDFTerm getBaseIRI(final RDFTerm object) {
...
+        return supplyAsync(() -> createOrReplace(identifier, ixnModel, dataset, OperationType.CREATE));
+    }
+
+    @Override
+    public Future<Boolean> delete(final IRI identifier, final IRI ixnModel, final Dataset dataset) {
+        LOGGER.debug("Deleting: {}", identifier);
         return supplyAsync(() -> {
-            final Boolean isDelete = dataset.contains(of(PreferAudit), null, RDF.type, AS.Delete);
             final Instant eventTime = now();
+            dataset.clear();
```
acoburn: Actually, this is a bug, and I take credit for introducing it. Simply removing this line should work. See related conversation here: https://github.com/trellis-ldp/trellis/pull/31#discussion_r169131736

ajs6f: You want me to pull out that line in this PR?

acoburn: Yes, please. It would have been in one of the commits I added to the PR. Alternately, we can let it stay and I'll open a separate issue to pull it out after the PR is merged.

ajs6f: Done.

trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -577,7 +587,9 @@ private void init() {
 
     @Override
     public Optional<Resource> get(final IRI identifier, final Instant time) {
-        return mementoService.map(svc -> svc.get(identifier, time)).orElseGet(() -> get(identifier));
+        // TODO -- JDK9 replace with Optional::or
+        final Optional<Resource> res = mementoService.flatMap(svc -> svc.get(identifier, time));
+        return res.isPresent() ? res : get(identifier);
```
acoburn: I am wondering about this function. Rather than checking `res.isPresent()`, I think a better approach would be:

```java
public Optional<Resource> get(final IRI identifier, final Instant time) {
    return mementoService.isPresent() ? mementoService.flatMap(svc -> svc.get(identifier, time)) : get(identifier);
}
```

This way, if the memento service isn't present, it defaults to a simple `::get`, but if the memento service _is_ present, and a client requests a datetime the precedes the creation of the resource, then an empty `Optional` is returned, which is what I'd expect (i.e. a 404 for requests that predate the resource). Right now the code returns the current state of the resource if no particular memento is found.

trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -141,58 +140,70 @@ private static RDFTerm getBaseIRI(final RDFTerm object) {
...
+                g.stream(identifier, LDP.isMemberOfRelation, null).findFirst().ifPresent(t -> dataset
+                                .add(PreferServerManaged, identifier, LDP.isMemberOfRelation, t.getObject()));
+                dataset.add(PreferServerManaged, identifier, LDP.insertedContentRelation,
+                                g.stream(identifier, LDP.insertedContentRelation, null).map(Triple::getObject)
+                                                .findFirst().orElse(LDP.MemberSubject));
+            });
+        }
+        // Set the parent relationship
+        getContainer(identifier).ifPresent(parent -> dataset.add(PreferServerManaged, identifier, DC.isPartOf, parent));
+        mementoService.ifPresent(svc -> get(identifier).ifPresent(res -> svc.put(identifier, eventTime, res.stream())));
```
acoburn: I've written some Memento tests, and I've run into an issue related to this line that will require a little thought.

Imagine this scenario:

1. An LDP-RS is created with some number of triples.
2. Some time passes, and the triples are changed.
3. Some more time passes and some of the triples are changed again.

Now, a simple GET request will fetch the current state of the resource, but how many Memento resources are there?

Given the code here in your branch: https://git.io/vAaMW there will be two Mementos.

Based on the code currently in master: https://git.io/vAaSv there will be three Mementos.

The difference, code-wise, has to do with when the `::get(identifier)` is called when passing data to the mementoService. In your version, that is called before the new data is stored, in the current master, it is called after the data is stored.

In another sense, the answer depends on whether a memento should be created immediately when a resource is created (that is, the latest Memento is always equivalent to the current state of the resource) or whether creation should happen only after the current state changes. My main consideration for the current behavior has to do with client expectations. That is, if the state of a resource at time X is X' and the state of the resource at time Y is Y', then a request for the resource at (X <= time < Y) should return resource state X'. Likewise, a request for the resource at (Y <= time) should return Y'. But the code in your PR offsets the Mementos. That is, if the resource is changed exactly at time Y, then a request for a Memento at time Y will return X'. Assume that the time precision is 1 second, due to HTTP spec limitations.

acoburn: Another thought on this and the issue above about the memento service -- both are related specifically to the triplestore impl and somewhat orthogonal to the main thrust of this PR. It's up to you, but it would be totally find to leave the implementation code as-is and deal with it separately. I certainly don't want to hold up this effort.

trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
ajs6f: WTF? I did this change (I agree with your point about the semantics of time-sensitive`::get`, but now the app tests are failing... can you see that? I was going to do this and merge, but hm...

acoburn: You'll want to revert this last change (i.e. leaving the time-sensitive `::get` as it was) -- there are a number of interrelated issues, beginning with some code in the HTTP layer, and it would be better to simply merge all of this and deal with those changes at once.

ajs6f: Can do. Let's land this darn thing (I have let it fly for way too long) and charge forward.

PR #53, opened 2018-02-27
adds version to banner and changes style

christopher-johnson:
a trivial change that has one function, to notify the console with a build version.  The style of the banner is bolder and reflects the solidity of the Trellis design.  Also, gradle will not process a resource with the special characters contained in the original version.


acoburn:
You don't like the flowers? :-)

PR #54, opened 2018-03-01
Add CDI support

### LINKED ISSUE #12: Add CDI support to HTTP (and other) classes
[CDI](https://docs.oracle.com/javaee/7/tutorial/cdi-basic.htm) offers a standard J2EE-based dependency injection framework that could greatly simplify object creation in Trellis.

acoburn:
Resolves #12

There are several changes here:

First, this adds an `@Inject` annotation to each CDI-managed bean (unless there is already a no-arg constructor).

Second, each package with CDI-managed beans has a `beans.xml` file in the appropriate resources directory.

Third, component configuration has been consolidated to use Apache Tamaya.

Fourth, constructors have been cleaned up to work better in CDI environments while remaining compatible with non-CDI environments.

Fifth, constructors that had been using generic types (e.g. the various cache-related types) now use concrete types. E.g.:

    CacheService<String, String>

becomes:

    ProfileCacheService

I will note that I am not actually testing this change in a CDI environment, but that will come. Once that happens, there may be additional changes, but this PR lays the foundation for all of that.


ajs6f:
Will def review, @ us2ts.org today/tomorrow-- can I review before Monday?

ajs6f:
First question: looks like the `beans.xml`s are all the same. Is the case that this PR makes all beans try to load all the time, and relies on module/JAR management to wire up the actual selections? (Not an objection, just trying to figure out how far this goes.)

acoburn:
@ajs6f my thinking here was just to rely on module/JAR management, but you raise a good point about the content of the beans.xml files. The current approach is definitely pretty simplistic, and if you have a suggestion for improving it, I'm all ears.

ajs6f:
I think it's legit for now-- step by step. For example, the next big step would be to introduce the use of actual CDI framework in `trellis-app` and do wiring there, perhaps via Tamaya-based config. But getting `beans.xml` into play is a good step and I have no objection.

trellis-webac/build.gradle
```diff
@@ -9,6 +9,7 @@ ext {
 dependencies {
 
     api group: 'org.apache.commons', name: 'commons-rdf-api', version: commonsRdfVersion
+    api group: 'org.apache.servicemix.bundles', name: 'org.apache.servicemix.bundles.javax-inject', version: javaxInjectVersion
```
ajs6f: Is it this (and not the vanilla API JAR) in order to get OSGi metadata?

ajs6f:
Is the use of concrete types specifically to make typesafe injection work without colliding caches?

trellis-webac/build.gradle
acoburn: Yes, it's just to make it easier to deploy in Karaf.

acoburn:
Yes, the concrete types are created specifically to make it easier to distinguish between the different caches, since those caches make use of different generic types and they also will typically have very different configurations. Another option would be to use annotations, but this seemed more straight-forward.

ajs6f:
Yeah, I probably would have gone with annotations, but I couldn't give you a good reason for it. But it might be good to comment on those new types that they exist for the purpose of injection and aren't intended to actually extend the generic types in any way.

acoburn:
@ajs6f I am going to update this PR with annotations rather than using specialized types for the various caches.

ajs6f:
I'm in no way against that, but just to be clear, I'm not demanding that. I think we're okay to go either way... but happy to review again once you've sprinkled the annotation flavor on top!

acoburn:
@ajs6f: yes, either way is fine with me, too. I just think the types would have a cleaner design with annotations.

ajs6f:
That's probably true. And they might be reusable if the types of caches change, which would be nice.

acoburn:
@ajs6f I have removed the concrete types in favor of qualifiers for use with CDI injection. I am using `@Named("...")` qualifiers, which is the simplest form possible. It would also be possible to define our own qualifier types, but for now this seemed to be the easiest to implement, and until we know exactly how this all will be used in a full CDI environment, this seems like the best way to start with qualifiers.

ajs6f:
I hate injection-by-name (nuts to Spring!) but I agree that it's a sensible move here. We can use this and work on our meta-annotation hype as we go.

I'm 👍 to this PR.

acoburn:
@ajs6f sounds good. We can always refine the qualifiers as we go, but this gets us a long ways there.

Issue #57, opened 2018-03-07
AS message produced without provenence / type


christopher-johnson:
I am working on a new camel-kafka-elasticsearch integration and have noticed a possible issue related with the triplestore resource service event implementation.  For some reason (as yet unknown) the event type (e.g. <https://www.w3.org/ns/activitystreams#Create>) is empty.  

Here is an example AS message sourced from Kafka for reference:
```json
{
  "@context" : "https://www.w3.org/ns/activitystreams",
  "id" : "urn:uuid:525573b5-f20f-489d-99dd-33668d3534bb",
  "type" : [ ],
  "object" : {
    "id" : "http://trellis:8080/ldp-test-6b56d629-3150-4ff5-8f69-1efa672f60fb",
    "type" : [ "http://www.w3.org/ns/ldp#RDFSource", "http://www.w3.org/ns/oa#TimeState", "http://www.w3.org/ns/activitystreams#Application", "http://xmlns.com/foaf/0.1/Person", "http://www.w3.org/ns/oa#Choice", "http://www.w3.org/ns/oa#HttpRequestState", "http://purl.org/dc/dcmitype/Sound", "http://www.w3.org/ns/oa#SpecificResource", "http://www.w3.org/ns/oa#TextualBody", "http://www.w3.org/ns/oa#Annotation", "http://www.w3.org/ns/oa#TextPositionSelector", "http://www.w3.org/ns/oa#FragmentSelector", "http://www.w3.org/ns/oa#CssStyle" ]
  },
  "published" : "2018-03-07T07:15:42.159590Z"
}
```
The target types are populated.  The triple is created.  Could have a simple explanation, I will keep looking at it.

christopher-johnson:
It seems that this could be related to reading method for the `PreferAudit` graph.  I notice that the AS message also does not include `actor`.   

Even though the audit triples are created in the `<http://www.trellisldp.org/ns/trellis#PreferAudit>` graph, a request with the `?ext=audit` does not return them (which I guess is the right way to do this, though the [documentation](https://github.com/trellis-ldp/trellis/wiki/Resource-Provenance--(Audit)) suggests alternatively `Prefer: return=representation; include="http://www.trellisldp.org/ns/trellis#PreferAudit"`  

If the `fetchAuditQuads` method cannot collect any data using `ext=audit`, then the message values would be empty.

acoburn:
[...]

PR #59, opened 2018-03-07
Add session information to API methods

### LINKED ISSUE #56: Add session information to mutating ResourceService methods
The mutating ResourceService methods (`::create`, `::replace` and `::delete`) will cause events to be emitted. Before the recent refactor in #31, this data had been available in the Audit-related triples, but now that data is only included in the `::add` method. By including a `Session` value (or similar) in the mutating methods, it would be possible to get access to these data.

This would also be an opportunity to remove the use of the default graph in these methods, the purpose of which relates only to providing data to the event producer. This likely will require adding a field (e.g. `baseURL`) to the `Session` type. This will also lead to some cleaner code in the HTTP layer.

acoburn:
Resolves #56 and #57

This updates the Trellis API ResourceService methods
by adding session-related data in a distinct type.

This fixes errors in the event stream where event type and agent
data were missing from messages.

### DIFF
--- a/trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
+++ b/trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
@@ -261,23 +261,11 @@ public class TriplestoreResourceService extends DefaultAuditService implements R
 
         eventService.ifPresent(svc -> {
             svc.emit(new SimpleEvent(getUrl(identifier, baseUrl),
-                        asList(session.getAgent()), asList(PROV.Activity, getIriFromOperationType(opType)),
+                        asList(session.getAgent()), asList(PROV.Activity, OperationType.asIRI(opType)),
                         targetTypes, inbox));
             getContainer(identifier).ifPresent(parent ->
                     emitEventsForAdjacentResources(svc, parent, session, opType, time));
         });
-    }
-
-    private IRI getIriFromOperationType(final OperationType opType) {
-        switch (opType) {
-            case DELETE:
-              return AS.Delete;
-            case CREATE:
-              return AS.Create;
-            case REPLACE:
-            default:
-              return AS.Update;
-        }
     }
 
     /**
@@ -423,7 +411,19 @@ public class TriplestoreResourceService extends DefaultAuditService implements R
     }
 
     private enum OperationType {
-        DELETE, CREATE, REPLACE
+        DELETE, CREATE, REPLACE;
+
+        static IRI asIRI(final OperationType opType) {
+            switch (opType) {
+                case DELETE:
+                  return AS.Delete;
+                case CREATE:
+                  return AS.Create;
+                case REPLACE:
+                default:
+                  return AS.Update;
+            }
+        }
     }
 
     /**
--- a/trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
+++ b/trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
@@ -660,7 +660,7 @@ public class TriplestoreResourceService extends DefaultAuditService implements R
     }
 
     @Override
-    public Future<Boolean> add(final IRI id, final Dataset dataset) {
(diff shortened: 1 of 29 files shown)


acoburn:
@ajs6f if you don't mind reviewing, that'd be great. One item in particular is that the MutableResource has three types, e.g. `<IRI, Session, Resource>` and the ImmutableResource has just two: `<IRI, Resource>`. There isn't presently any need to pass `Session` information into the ImmutableResource::add method, but it does introduce an asymmetry. Do you think that the ImmutableResource type should also change to `ImmutableResource<T, U, V>` or should it stay as `ImmutableResource<T, U>`?

ajs6f:
What would be the use case for a `MutableDataService<T, U, V>` where `U` is anything other than the current Trellis `Session` type?

Another way to think about it; rather than extend the type signature of those services, maybe we should DEgenerify them entirely unless/until we actually have a use case for something that say, stores mutable data under some key that _isn't_ an `IRI`, etc.

I don't remember exactly why I wrote them generically to begin with. Maybe I thought they could be used in other contexts than the main persistence path? Anyway, I can't think of a reason now...

acoburn:
I don't know of a reason that the type wouldn't be a `Session`. I see no reason why the type signature couldn't be just `MutableDataService<T, U>` with a concrete `Session` type in the methods.

ajs6f:
Simple != good, but IME generally simple => good.

acoburn:
I could continue to de-generify those interfaces by setting the first type as an `IRI` if you think that would be a good idea; I don't have a strong opinion one way or the other.

ajs6f:
Let's do it in a separate PR (for history's sake) and I volunteer for that once this one has landed. It'll be pretty darn easy.

ajs6f:
Okay, thought about it for a bit and I say that we _should_ add `Session` to the immutable side as well. I can think of decently real use cases in which one might like to record audit-ish info _including_ the base URL.

trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -231,20 +241,45 @@ private Boolean storeAndNotify(final IRI identifier, final Dataset dataset, fina
...
+            svc.emit(new SimpleEvent(getUrl(identifier, baseUrl),
+                        asList(session.getAgent()), asList(PROV.Activity, getIriFromOperationType(opType)),
+                        targetTypes, inbox));
             getContainer(identifier).ifPresent(parent ->
-                    emitEventsForAdjacentResources(svc, parent, time, baseUrl, dataset));
+                    emitEventsForAdjacentResources(svc, parent, session, opType, time));
         });
     }
 
+    private IRI getIriFromOperationType(final OperationType opType) {
```
ajs6f: Should we put this in `OperationType` (as a `public` constant or as a `static` method in the `enum`)?

acoburn: Yes, that makes sense to me.

trellis-triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -261,11 +284,14 @@ private void emitEvents(final IRI identifier, final Literal time, final Dataset
...
+     * <p>Note: this query does not retrieve the user-managed rdf:type values
+     * nor does it retrieve any ldp:inbox value.
      */
-    private void emitEventsForAdjacentResources(final EventService svc, final IRI parent,
-            final Literal time, final Optional<String> baseUrl, final Dataset dataset) {
-        final Boolean isDelete = dataset.contains(of(PreferAudit), null, RDF.type, AS.Delete);
-        final Boolean isCreate = dataset.contains(of(PreferAudit), null, RDF.type, AS.Create);
+    private void emitEventsForAdjacentResources(final EventService svc, final IRI parent, final Session session,
+            final OperationType opType, final Literal time) {
+        final Boolean isDelete = opType == OperationType.DELETE;
```
ajs6f: Just a note for the future-- this is a smell. We'll want to introduce some polymorphism at some point to avoid this kind of switchery. But I'm not asking for that now.

PR #60, opened 2018-03-09
*Service<T,U> => *Service<U> => *Service<T>

ajs6f:


acoburn:
Nice! Thanks so much.

ajs6f:
I shouldn't have made them so generic to begin with. 🤷‍♂️

Issue #58, opened 2018-03-07
HTTP/2 over TLS  support requires ALPN dependency


christopher-johnson:
In order to make trellis-app with dropwizard's h2 support configuration in JDK 9 or JDK 10, I added a dependency:
```groovy
compile group: 'org.eclipse.jetty', name: 'jetty-alpn-java-server', version: '9.4.8.v20171121'
```
This version is built for JDK 9, but it works on JDK 10.

acoburn:
@christopher-johnson would make sense to add a profile that selectively adds this dependency? I would want to make sure that h2 is still well supported on JDK 1.8

christopher-johnson:
In order to make ALPN work on JDK 8, one needs to add `-Xbootclasspath/p:$path_to_alpn_boot.jar` as a JVM option.  This is probably not feasible to do with gradle as it involves copying the alpn_boot library (from some flat_dir in the repo) to the build host, and then adding a `applicationDefaultJvmArgs` to modify the application start script.  It is pretty easy to set `JAVA_OPTS` in docker, and then copy the lib into the image at that path, with something like this:
```yaml
ENV ALPN_BOOT alpn-boot-8.1.12.v20180117.jar
ENV TRELLIS_RUNTIME /opt
ENV JAVA_OPTS="-Xms250m -Xmx1024m -Xbootclasspath/p:${TRELLIS_RUNTIME}/${ALPN_BOOT}"
```
I tested this and it works.

The question of how or if to conditionally support 9 ,10 and eventually 11 builds in the trellis gradle scripts is open.  To build trellis-app with H2 for 9 only requires a couple of changes (adding the compile dependency on alpn-java-server and changing the target to 1.9).   The alpn-java-server is a transitive dependency that is already included with dropwizard http2, and I guess the gradle declaration includes it to the runtime classpath(?).

To build for 10 requires excluding jacoco (you know about this), and to avoid javadoc warnings, adding this:
```groovy
task docs(type: Javadoc) {
        doFirst {
            if (Jvm.current().getJavaVersion().isJava10Compatible()) {
                addBooleanOption('html5', true)
            }
        }
    ...
```   
One other note is that the trellis-app `build.gradle` includes this:
```groovy
artifacts {
    archives javadocJar
    archives sourceJar
}
```
I think that this can be removed since it is already in the rootProject build.gradle.

acoburn:
@christopher-johnson feel free to issue a PR that removes the redundant `artifacts` section.

[...]

PR #63, opened 2018-03-11
removes redundant artifacts

christopher-johnson:
See https://github.com/trellis-ldp/trellis/issues/58#issuecomment-372062967


acoburn:
Thanks, @christopher-johnson!

PR #74, opened 2018-03-20
Factoring apart component builds for easier overriding

ajs6f:
This commit introduces `protected` methods for most of the parts of a running Trellis instance, in order to make subclassing `TrellisApplication` less filled with cut-and-paste.

### DIFF
--- a/trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
+++ b/trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
@@ -74,7 +74,7 @@ public class TrellisApplication extends Application<TrellisConfiguration> {
 
     private MementoService mementoService;
 
-    private TriplestoreResourceService resourceService;
+    private ResourceService resourceService;
 
     private NamespaceService namespaceService;
 
@@ -157,7 +157,7 @@ public class TrellisApplication extends Application<TrellisConfiguration> {
 
     protected AuditService buildAuditService() {
         return resourceService != null && resourceService instanceof AuditService
-                        ? resourceService
+                        ? (AuditService) resourceService
                         : new DefaultAuditService();
     }
 
--- a/trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
+++ b/trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
@@ -22,12 +22,12 @@ import static org.trellisldp.app.TrellisUtils.getNotificationService;
 import static org.trellisldp.app.TrellisUtils.getRDFConnection;
 import static org.trellisldp.app.TrellisUtils.getWebacConfiguration;
 
-import javax.jms.JMSException;
-
 import io.dropwizard.Application;
 import io.dropwizard.auth.chained.ChainedAuthFilter;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
+
+import javax.jms.JMSException;
 
 import org.apache.jena.rdfconnection.RDFConnection;
 import org.trellisldp.agent.SimpleAgent;
--- a/trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
+++ b/trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
@@ -22,6 +22,8 @@ import static org.trellisldp.app.TrellisUtils.getNotificationService;
 import static org.trellisldp.app.TrellisUtils.getRDFConnection;
 import static org.trellisldp.app.TrellisUtils.getWebacConfiguration;
 
+import com.google.common.cache.Cache;
+
 import io.dropwizard.Application;
 import io.dropwizard.auth.chained.ChainedAuthFilter;
 import io.dropwizard.setup.Bootstrap;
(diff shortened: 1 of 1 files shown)


ajs6f:
This PR will fail to build because `checkstyle` does not like the import order. I tried every permutation I could think of and couldn't get it to work.

Does anyone who uses Eclipse have a formatter configuration that agrees with whatever Checkstyle spec is in use?

christopher-johnson:
the imports have to be alphabetical.  So `import javax.jms.JMSException` would be after `import io.dropwizard.Application`

trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
```diff
@@ -22,6 +22,8 @@
 import static org.trellisldp.app.TrellisUtils.getRDFConnection;
 import static org.trellisldp.app.TrellisUtils.getWebacConfiguration;
 
+import javax.jms.JMSException;
```
acoburn: This should follow the `import io....` section

ajs6f:
Thanks, guys.

acoburn:
The only issue here is that the `TriplestoreResourceService` implements both the `AuditService` and `ResourceService` interfaces, and a different persistence impl may separate those things. Instead, I wonder if there shouldn't rather be some sort of shared `init()` (or `PostConstruct`-annotated) method and the protected methods are more like: `get*Service()` instead of `build*Service`. That way, if some of those objects implement multiple interfaces, there wouldn't be an issue.

ajs6f:
Oh, shoot, you're right, and what's more, I think I did that. Let me do a little quick juggling and see if I can keep this PR simple, and yes, I'll try to use some fields and an init for this. That sounds good. The next commit might be a little while coming because the storm is getting worse here and I might want to get myself home before digging into this.

ajs6f:
Looks like the JDK10 Travis build failed for unrelated reasons (couldn't download the JDK).

acoburn:
Don't worry about the JDK 10 build -- I figured that the download URL would change today. I'll fix that separately.

acoburn:
The JDK 10 download issue has been resolved: https://github.com/trellis-ldp/trellis/commit/4ceeb231b502b37395c32fc9974f2aaa30f658f0

ajs6f:
Better?

trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
```diff
@@ -56,6 +64,28 @@
...
+
+    private TrellisConfiguration config;
+
+    private EventService notificationService;
+
+    private IdentifierService idService;
+
+    private MementoService mementoService;
+
+    private TriplestoreResourceService resourceService;
```
acoburn: Presumably, a subclass would not use this field? In other words, doesn't this add a hard dependency on the Triplestore implementation?

ajs6f:
Arg! Fixed. My excuse is that I just got home from the snow.

trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
```diff
@@ -127,4 +146,55 @@ public void run(final TrellisConfiguration config,
...
+    }
+
+    protected TrellisCache<String, String> buildCacheService() {
+        final Long cacheSize = config.getJsonld().getCacheSize();
+        final Long hours = config.getJsonld().getCacheExpireHours();
+        final Cache<String, String> cache = newBuilder().maximumSize(cacheSize).expireAfterAccess(hours, HOURS).build();
+        return new TrellisCache<>(cache);
+    }
+
+    protected JenaIOService buildIoService(final NamespaceService namespaceService,
```
acoburn: `JenaIOService` => `IOService`.

trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
```diff
@@ -127,4 +146,55 @@ public void run(final TrellisConfiguration config,
...
+                    final CacheService<String, String> profileCache) {
+        return new JenaIOService(namespaceService, profileCache, TrellisUtils.getAssetConfiguration(config));
+    }
+
+    protected FileBinaryService buildBinaryService(final IdentifierService idService) {
```
acoburn: `FileBinaryService` => `BinaryService`

trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
```diff
@@ -127,4 +146,55 @@ public void run(final TrellisConfiguration config,
...
+        return new FileBinaryService(idService, config.getBinaries(), config.getBinaryHierarchyLevels(),
+                        config.getBinaryHierarchyLength());
+    }
+
+    protected NamespacesJsonContext buildNamespaceService() {
```
acoburn: `NamespacesJsonContext` => `NamespaceService`

trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
```diff
@@ -127,4 +146,55 @@ public void run(final TrellisConfiguration config,
...
+        return new NamespacesJsonContext(config.getNamespaces());
+    }
+
+    protected FileMementoService buildMementoService() {
```
acoburn: `FileMementoService` => `MementoService`

trellis-app/src/main/java/org/trellisldp/app/TrellisApplication.java
```diff
@@ -127,4 +146,55 @@ public void run(final TrellisConfiguration config,
...
+        return new FileMementoService(config.getMementos());
+    }
+
+    protected UUIDGenerator buildIdService() {
```
acoburn: `UUIDGenerator` => `IdentifierService`

ajs6f:
Arg. Sorry, doing too much at once.

PR #79, opened 2018-03-21
Offering extended configuration

ajs6f:
Adds configuration for use with [Trellis-Cassandra](https://github.com/ajs6f/trellis-cassandra).

### DIFF
--- a/trellis-app/src/main/java/org/trellisldp/app/config/TrellisConfiguration.java
+++ b/trellis-app/src/main/java/org/trellisldp/app/config/TrellisConfiguration.java
@@ -13,9 +13,17 @@
  */
 package org.trellisldp.app.config;
 
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import io.dropwizard.Configuration;
+
+import static java.util.Collections.synchronizedMap;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
 import javax.validation.constraints.NotNull;
 
@@ -62,10 +70,8 @@ public class TrellisConfiguration extends Configuration {
     private String baseUrl = null;
 
     private String resourceLocation = null;
-
-    private String cassandraAddress = "localhost";
-
-    private Integer cassandraPort = 9042;
+    
+    private final Map<String, Object> extras = synchronizedMap(new HashMap<>());
 
     /**
      * Get the base URL for the partition.
@@ -176,39 +182,24 @@ public class TrellisConfiguration extends Configuration {
     }
 
     /**
-     * Set the Cassandra cluster address.
-     * @param addy the Cassandra cluster address
+     * Set an extra configuration value.
+     * @param name the name of this config value
+     * @param value the value to set
+     * @return this config for chaining
      */
-    @JsonProperty
-    public void setCassandraAddress(final String addy) {
-        this.cassandraAddress = addy;
+    @JsonAnySetter
+    public TrellisConfiguration setAdditionalConfig(final String name, final Object value) {
+        extras.put(name, value);
+        return this;
     }
 
     /**
-     * Get the Cassandra cluster address.
-     * @return the Cassandra cluster address
+     * Get any extra metadata.
+     * @return a {@link Map} of any extra metadata
      */
-    @JsonProperty
-    public String getCassandraAddress() {
-        return cassandraAddress;
-    }
-
-    /**
-     * Set the Cassandra cluster port.
-     * @param port the Cassandra cluster port
-     */
-    @JsonProperty
-    public void setCassandraPort(final int port) {
-        this.cassandraPort = port;
-    }
-
-    /**
-     * Get the Cassandra cluster port.
-     * @return the Cassandra cluster port
-     */
-    @JsonProperty
-    public Integer getCassandraPort() {
-        return cassandraPort;
+    @JsonAnyGetter
+    public Map<String, Object> any() {
+        return extras;
     }
 
     /**
--- a/trellis-app/src/main/java/org/trellisldp/app/config/TrellisConfiguration.java
+++ b/trellis-app/src/main/java/org/trellisldp/app/config/TrellisConfiguration.java
@@ -13,15 +13,14 @@
  */
 package org.trellisldp.app.config;
 
+import static java.util.Collections.synchronizedMap;
+
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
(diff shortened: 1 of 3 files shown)


acoburn:
This seems fine, but I wonder if it wouldn't be better for the trellis-over-cassandra-app to subclass `TrellisConfiguration` and add these methods there?

ajs6f:
You tell me-- I know as much about Dropwizard as I do about playing poker: nothing. Should I close this and try that instead?

ajs6f:
Or do I just cast the config? Is it legit to:
```

```

ajs6f:
Sorry-- misclick.

acoburn:
A better approach might be to add a `@JsonAnySetter` method to `TrellisConfiguration`:

```java
@JsonAnySetter
public void set(final String name, final String value) {
    internalMap.put(name, value);
}

@JsonAnyGetter
public Map<String, String> any() {
    return internalMap;
}
```

That adds a simple extension point for any possible subclasses.

ajs6f:
Where is `internalMap`? Is that to be added as well?

ajs6f:
I'm not much with JSON; what about `Map<String, Object>`? Would that effectively account for nesting?

acoburn:
I haven't ever used that precise construct with Jackson serializations, but I don't see why it wouldn't work. It should be trivial to set up a test -- just add some additional fields to one of the configuration files in `./src/test/resources`

ajs6f:
Didn't work at all, not even with simple `String`s:
```
io.dropwizard.configuration.ConfigurationParsingException: /Users/ajs6f/Documents/trellis/trellis/trellis-app/bin/test/config1.yml has an error:
  * Unrecognized field at: cassandraAddress
    Did you mean?:
      - assets
      - resources
      - namespaces
      - cacheMaxAge
      - binaries
        [12 more]
```

So maybe that works for Jackson's semantics, but not for Dropwizard's particular concrete processing code?

ajs6f:
Again, I know little about Dropwizard, but I'm beginning to suspect that we're running up against its intentional limits-- it seems to be for building single apps, not ecosystems. Extensible config is really not a concern if you are build a single app.

ajs6f:
Okay, how's that?

ajs6f:
incidentally, I tried going:
```
public class TrellisApplication extends Application<TrellisConfiguration> {

```
=>
```
public class TrellisApplication<T extends TrellisConfiguration> extends Application<T> {

```
but all of the test classes exploded with:
```
The constructor DropwizardTestSupport<TrellisConfiguration>(Class<TrellisApplication>, String, ConfigOverride, ConfigOverride, ConfigOverride) is undefined
```

Not really sure what's going on there, but I _think_ the issue is that `Class<TrellisApplication>` isn't provably a supertype of `Class<TrellisApplication<T extends TrellisConfiguration>>` (or the other way 'round), so the compiler pukes into its own shirt. Shall we just rewrite the whole app in Scala?

christopher-johnson:
@ajs6f I tried running the test and it works for me using IntelliJ.  One side note is that the `@RunWith(JUnitPlatform.class)` is not required since the test class is written for JUnit5.  AFAIK, that annotation invokes the JUnit4 runner.

ajs6f:
Thanks much @christopher-johnson ! I didn't actually write that test, just added a few limes to it, so I'll let @acoburn remove that annotation. Unless, @acoburn , you'd like me to do it here?

trellis-app/src/main/java/org/trellisldp/app/config/TrellisConfiguration.java
```diff
@@ -13,10 +13,18 @@
...
 
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import io.dropwizard.Configuration;
 
+import static java.util.Collections.synchronizedMap;
+
+import java.util.Collections;
```
acoburn: This isn't used and can be removed.

trellis-app/src/main/java/org/trellisldp/app/config/TrellisConfiguration.java
```diff
@@ -13,10 +13,18 @@
  */
 package org.trellisldp.app.config;
 
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import io.dropwizard.Configuration;
 
+import static java.util.Collections.synchronizedMap;
```
acoburn: Move `static import`s to the top (before ordinary imports)

trellis-app/src/main/java/org/trellisldp/app/config/TrellisConfiguration.java
```diff
@@ -62,6 +70,8 @@
     private String baseUrl = null;
 
     private String resourceLocation = null;
+
```
acoburn: trailing whitespace

trellis-app/src/test/java/org/trellisldp/app/config/TrellisConfigurationTest.java
```diff
@@ -49,6 +52,14 @@ public void testConfigurationGeneral1() throws Exception {
         assertNull(config.getResources());
         assertEquals((Integer) 2, config.getBinaryHierarchyLevels());
         assertEquals((Integer) 1, config.getBinaryHierarchyLength());
+        assertEquals("my.cluster.node", config.any().get("cassandraAddress"));
+        assertEquals((Integer)245993, config.any().get("cassandraPort"));
+        @SuppressWarnings("unchecked")
+        Map<String, Object> extraConfig = (Map<String, Object>) config.any().get("extraConfigValues");
```
acoburn: add `final` here.

trellis-app/src/test/java/org/trellisldp/app/config/TrellisConfigurationTest.java
```diff
@@ -49,6 +52,14 @@ public void testConfigurationGeneral1() throws Exception {
...
+        assertTrue((Boolean) extraConfig.get("one"));
+        @SuppressWarnings("unchecked")
+        List<String> list = (List<String>) extraConfig.get("two");
```
acoburn: add `final`

ajs6f:
I will fix those probs in two seconds, but just so you know-- I ran `./gradlew check install` to check this and it passed fine. Do we need to couple `checkstyleMain` into `check`? I thought it already was?

acoburn:
That's odd. It could be that running this without `clean` means that the changes aren't caught. I definitely get the checkstyle errors when running `./gradlew clean check install`

ajs6f:
Okay, I'll keep an eye peeled. I might try making a few intentional faux pas and see which Gradle invocations catch them and which do not.

Issue #80, opened 2018-03-21
Problem with getting absolute paths


ajs6f:
`BaseCommonTests` features a method 
```
    protected static String getResourcePath(final String path) {
        try {
            return new File(BaseCommonTests.class.getResource(path).toURI()).getAbsolutePath();
        } catch (final URISyntaxException ex) {
            throw new RuntimeException(ex);
        }
    }
```
This is really unstable and fails every time under Maven with:
```
Caused by: java.lang.IllegalArgumentException: URI is not hierarchical
	at java.io.File.<init>(File.java:418)
	at org.trellisldp.test.BaseCommonTests.getResourcePath(BaseCommonTests.java:104)
	at [somewhere that called getResourcePath]
```
presumably because `BaseCommonTests.class.getResource(path).toURI()` is indeed not returning an URI that can be used for a file. In fact, I think this might only work in Gradle.

I'm not sure what to do here: ideally, we would decouple fully from build mechanics for tests, because I'm trying to reuse tests for `trellis-cassandra-app` and this is blocking me. But I certainly don't want to build up a whole resource management machinery instead!

In similar situations in Maven, I've passed in a system property with the absolute build directory. I wouldn't claim that so doing is that much better.

ajs6f:
Some more details: it looks like this works on Gradle because Gradle is finding the path of the copy of a file that has been placed in the `bin/` directory, whereas Maven is looking inside the actual JAR that has been built out of resources. I'm going to send a PR for overriding the problematic method-- it's the best thing I can think of in the circumstances, but a more elegant option is almost a desideratum.

PR #81, opened 2018-03-21
Make getResourcePath overridable for build systems other than Gradle

ajs6f:
Fixes https://github.com/trellis-ldp/trellis/issues/80


ajs6f:
I added one commit that (as the comment says) removes a hard dependency on Tamaya. That was also blowing up Maven. I can cherrypick that off to another PR if desired.

With this PR and https://github.com/trellis-ldp/trellis/pull/79, `trellis-cassandra-app` can pass the LDP tests from `trellis-test`! Yay!

PR #82, opened 2018-03-22
Introducing NoopNamespaceService!

ajs6f:


acoburn:
This is great. What do you think: would it be better for this to go into the `trellis-api` module (where we already have a `NoopEventService` and a `NoopAuditService`)?

ajs6f:
Yeah, I was going to ask you-- I think it might be better there, because then people who don't want server-side ns management don't need to depend on `trellis-namespace`. I'll move it.

ajs6f:
I rewrote the commit to move it.

trellis-api/src/main/java/org/trellisldp/api/NoopNamespaceService.java
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+package org.trellisldp.api;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+
+import org.trellisldp.api.NamespaceService;
```
acoburn: You should remove this import. (redundant import from the same package)

trellis-api/src/test/java/org/trellisldp/api/NoopNamespaceServiceTest.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.trellisldp.api;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+import org.trellisldp.api.NamespaceService;
```
acoburn: Ditto. you can remove this import.

trellis-api/src/test/java/org/trellisldp/api/NoopNamespaceServiceTest.java
```diff
@@ -0,0 +1,32 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+public class NoopNamespaceServiceTest {
+
+    @Test
+    public void noAction() {
+        NamespaceService testService = new NoopNamespaceService();
```
acoburn: You should add `final` here.

ajs6f: Ah, yes, but also-- the reason you are catching these things and I am not is because my build keeps passing locally. So something is off about my Gradle setup, because it shouldn't be passing this, should it?

It should be running both `checkstyleMain` and `checkstyleTest` (which is what would have gotten this) for the `check` task, right?

acoburn: Correct. Are you running the gradle build from within your IDE or from the command line? And when building, are you running the command at the project root or in a subdirectory?

ajs6f: Always CLI, always from project root. Sometimes I will only build over a subdirectory/module (i.e. `gradlew trellis-api:checkstyleMain`).

PR #94, opened 2018-04-06
Move OSGi tests into their own module

### LINKED ISSUE #93: Build cycle
I'm not sure when this showed up (haven't looked at the code in a few days) but I'm getting a weird Eclipse error:
```
A cycle was detected in the build path of project 'trellis-test'. The cycle consists of projects {trellis-test, trellis-triplestore}
```
(of course it shows up for `-triplestore` too). The CLI `./gradlew` build works fine. It appears to be a dependency cross via the `testCompile` scope. Is this considered a problem in Gradle builds? I really have no idea.

acoburn:
Resolves #93

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -20,6 +20,7 @@ include ':trellis-webac'
 include ':trellis-app'
 
 include ':trellis-karaf'
+include ':trellis-osgi'
 include ':trellis-test'
 
 project(':trellis-api').projectDir = "$rootDir/trellis-api" as File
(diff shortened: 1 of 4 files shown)


acoburn:
Feel free to just merge it now (on a visual survey), or if you'd like to test it in Eclipse to see if that same error is generated, that would be ideal.

ajs6f:
Duh, why didn't I just do that first? Testing...

Issue #88, opened 2018-04-03
Factor out tests for nonBasic containers


ajs6f:
The current `org.trellisldp.test.ResourceServiceTests` tests the behavior of nonBasic container types like DC and IC. I haven't implemented them in `trellis-cassandra` yet, so these tests are failing. I can override those test methods for now, but it might be better at some point to factor them out to their own class.

acoburn:
Would it make sense to use Junit's [Conditional Test Execution](https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution) mechanism for this?

ajs6f:
Could be. The question I would ask is whether there is a solid programmatic way to understand whether a given impl supports nonBasic containers. E.g. some kind of characterizing method, or potentially a marker type or the like.

acoburn:
I don't believe that LDP provides a mechanism for discovering that -- other than creating a resource with the defined type and verifying that the request succeeds. And Trellis doesn't have a mechanism for that at present. Though perhaps the `ResourceService` interface could have a new method returning a list of supported interaction models.

ajs6f:
Yeah, I meant something Trellis-specific (in the Java). Absent some characterizing technique like the method you propose, it's not clear to me how conditional test execution knows when to run the extra tests....?

acoburn:
I suspect the cleanest approach would include an addition to the JAVA API. The current HTTP layer could even make use of that to control for "valid interaction models". Otherwise, in the testing code one would have to depend on system properties or the like (yuck!)

ajs6f:
One potential niceness might be earlier failing of requests with unsupported interaction models, before execution gets all the way to trying to persist.

PR #95, opened 2018-04-06
Add ResourceService::supportedInteractionModels method

### LINKED ISSUE #92: Add a method on ResourceService listing the supported interaction models
As discussed in #88, it would be helpful to introduce a method on the `ResourceService` interface that returns the supported LDP interaction models. That way, failures for non-supported LDP types can happen higher in the stack. It also makes testing more flexible, as non-relevant tests can be ignored.

One possibility for the signature for this method could be:

```java
Set<IRI> ResourceService::supportedInteractionModels();
```
 
The default implementation could include all LDP resource types (LDP-RS, LDP-NR, LDP-C, LDP-BC, LDP-DC, LDP-IC) but if a persistence layer has a different level of support, the method can be overridden to list only those types that are supported.

acoburn:
Resolves #92 

Once we have agreement on this approach, I can apply it to #88, allowing the test suite to adjust to varying levels of LDP container support.

I decided _not_ to include a default implementation in the interface -- if we want implementations to be clear about what is supported, then I thought it makes more sense to simply require an implementing class to _implement_ the method.

I am also making use of a new named individual in the Trellis ontology `trellis:InvalidInteractionModel`, which is used with the `ldp:constrainedBy` link header in the case of an error condition.

### DIFF
--- a/trellis-http/src/test/java/org/trellisldp/http/AbstractLdpResourceTest.java
+++ b/trellis-http/src/test/java/org/trellisldp/http/AbstractLdpResourceTest.java
@@ -13,6 +13,7 @@
  */
 package org.trellisldp.http;
 
+import static com.google.common.collect.Sets.newHashSet;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.time.Instant.MAX;
 import static java.time.Instant.ofEpochSecond;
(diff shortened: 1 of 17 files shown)


trellis-http/src/test/java/org/trellisldp/http/AbstractLdpResourceTest.java
```diff
@@ -200,6 +201,14 @@
         allModes.add(ACL.Control);
         allModes.add(ACL.Read);
         allModes.add(ACL.Write);
+
+        allInteractionModels.add(LDP.Resource);
```
ajs6f: If we have Guava on line (I think we do) maybe use something from `Sets` here to construct and fill the set?

trellis-http/src/test/java/org/trellisldp/http/impl/PostHandlerTest.java
```diff
@@ -88,6 +92,18 @@
...
+    private static final Set<IRI> allInteractionModels = new HashSet<>();
+
+    static {
+        allInteractionModels.add(LDP.Resource);
+        allInteractionModels.add(LDP.RDFSource);
+        allInteractionModels.add(LDP.NonRDFSource);
+        allInteractionModels.add(LDP.Container);
+        allInteractionModels.add(LDP.BasicContainer);
+        allInteractionModels.add(LDP.DirectContainer);
+        allInteractionModels.add(LDP.IndirectContainer);
```
ajs6f: See above about `Sets`.

trellis-http/src/test/java/org/trellisldp/http/impl/PutHandlerTest.java
```diff
@@ -94,6 +98,18 @@
...
+    private static final Set<IRI> allInteractionModels = new HashSet<>();
+
+    static {
+        allInteractionModels.add(LDP.Resource);
+        allInteractionModels.add(LDP.RDFSource);
+        allInteractionModels.add(LDP.NonRDFSource);
+        allInteractionModels.add(LDP.Container);
+        allInteractionModels.add(LDP.BasicContainer);
+        allInteractionModels.add(LDP.DirectContainer);
+        allInteractionModels.add(LDP.IndirectContainer);
```
ajs6f: Ditto.

trellis-http/src/test/java/org/trellisldp/http/AbstractLdpResourceTest.java
acoburn: Guava isn't a `compile` dependency, but I have no problem with adding it as a test dependency (if it's not already there).

acoburn:
I think the Trellis ontology should change `trellis:InvalidInteractionModel` to `trellis:UnsupportedInteractionModel`, which more accurately captures what is going on. I can update that as part of this PR or separately. WDYT?

ajs6f:
Just waiting for Travis et al to complete

PR #97, opened 2018-04-07
resolves #96

### LINKED ISSUE #96: JUnitPlatform causes ClassCastException in Gradle testCompile with JPMS
You may recall that I pointed out [here](https://github.com/trellis-ldp/trellis/pull/79#issuecomment-375083562) about removing the `@RunWith(JUnitPlatform.class)` annotation from the tests.  In doing an analysis of the `ClassCastException` that resulted from running any tests in the JPMS build, I discovered that this problem had nothing to do with the Logger or Gradle, but was actually caused by JUnit4.  

Check out this [passing build](https://travis-ci.org/trellis-ldp/trellis/builds/363478359) and [this commit](https://github.com/trellis-ldp/trellis/commit/797a71abaacc10bdf35589ed3495cccb58100f16) for how to fix this.  I think that it is safe to remove the annotation and the dependencies on the `junit-platform-runner`, but perhaps you want to verify.  Also, there are a few references [like this one](https://github.com/trellis-ldp/trellis/compare/jpms?expand=1#diff-b8a593b35e8912b2df16329cfaa35196) to JUnit4 methods that can be replaced

If you approve, then let me know if you need a PR.  The good thing is that all of the tests are now runnable in the JPMS branch!  I will continue to periodically merge master to keep this in sync.

christopher-johnson:
removes JUnit4 dependencies

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,12 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'org.junit.platform:junit-platform-gradle-plugin:1.1.1'
+    }
+}
+
 plugins {
     id 'com.github.hierynomus.license' version '0.14.0'
     id 'com.github.ben-manes.versions' version '0.17.0'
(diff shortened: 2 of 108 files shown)


acoburn:
From what I can tell, without that `@RunWith` annotation, the tests are not run at all.

christopher-johnson:
this could be a question of your IDE.  The tests definitely run for me without it.

acoburn:
`./gradle clean check jacocoRootReport` from the command line

christopher-johnson:
ok.  this is just a gradle configuration issue.  you need to add this 

```groovy
buildscript {
    repositories {
        mavenCentral()
    }
    dependencies {
        classpath 'org.junit.platform:junit-platform-gradle-plugin:1.1.0'
    }
}
```
and then add the runner as a dependency, like this:

```groovy
    dependencies {
        testCompile(
                'org.junit.jupiter:junit-jupiter-api:5.1.0'
        )

        testRuntime(
                'org.junit.jupiter:junit-jupiter-engine:5.1.0',
        )
    }
```
the tests then can be run from the CLI

acoburn:
Realistically, this PR will break all of the CI testing infrastructure (see [here](https://sonarcloud.io/dashboard?id=org.trellisldp%3Atrellis) and [here](https://coveralls.io/github/trellis-ldp/trellis)), because the JUnit runner is not part of the included project build script. I'd suggest either modifying the root project build script as above (so that the tests are run in all environments -- especially CI environments) or holding off on this until, as you say, gradle figures out the JUnit5 runner issue.

christopher-johnson:
There is an easier way to add JUnit 5 to gradle support since 4.6.  Check the [release notes](https://docs.gradle.org/4.6/release-notes.html)

You can add this to the test config:
```
        useJUnitPlatform {
            includeEngines 'junit-jupiter'
        }
```
Maybe try this and see if it works for you.  I notice some test failures now that I had not seen before...

acoburn:
This now looks great.

Two comments -- the stock JDK 10 root certs do not contain the cacerts for AWS, which is where the Trellis vocabularies are stored. You can see how this is handled for CI testing:

https://github.com/trellis-ldp/trellis/blob/master/buildtools/src/install/install-jdk-10.sh#L18-L26

Secondly, the `junit.platform.gradle.plugin` is now deprecated in favor of the `test.useJUnitPlatform` configuration. I'll send a PR on top of this one with that change.

acoburn:
@christopher-johnson do you know why the `trellis-osgi` tests don't run with this change? They are currently configured not to run under JDK 10 -- I'd like to change that now that Karaf 4.2.0 has been released -- but with this change those tests don't run at all under JDK8 or JDK9 either. And on the current master branch, the tests definitely do run (under JDK8/9).

christopher-johnson:
Yes.  I noted that [here](https://github.com/trellis-ldp/trellis/issues/96#issuecomment-379472553).

PaxExam still uses the JUnit4 `@RunWith` annotation.  This would require adding a `testRuntime` dependency in that build.gradle on the vintage engine:

```groovy
    testRuntime("org.junit.vintage:junit-vintage-engine:${junitVersion}")
```
I tested this locally and it works.  I push the change shortly

PR #108, opened 2018-04-13
Generalize trellis-app component

### LINKED ISSUE #99: Generalize trellis-app components
The `trellis-app` module has a hard dependency on `trellis-triplestore`, so re-using the `trellis-app` code for a different persistence layer isn't ideal. It would be better to factor all common elements into a `trellis-app-common` module -- one that has very few dependencies and then wire those together.

For example, much of what makes up the current `TrellisApplication` could be made into an abstract base class with some functions that need to be overridden, such as:

```java
abstract public Optional<ResourceService> getResourceService();

abstract public Optional<AuditService> getAuditService();

abstract public Optional<EventService> getEventService();
```

By using `Optional<T>` types, the base implementation code can select a no-op version in the absence of a wired-in service. For instance:

```java
final EventService eventService = getEventService.orElseGet(NoopEventService::new);
```

### LINKED ISSUE #104: stop pushing app distribution artifacts to maven central
The current build configuration pushes both a `.zip` and a `.tar` artifact to the maven repository. It would be better to omit these artifacts and make them available by other means, such as direct download.

One option would be to put all of the distribution-related code into its own gradle subproject, e.g. `./platform/distribution`

acoburn:
Resolves #99
Resolves #104 

This generalizes the dropwizard parts of `trellis-app` in order to make the component more reusable. That means that it does not compile-depend on the triplestore implementation (or any other concrete components -- other than WebAC). It also means that an implementing application needs only to extend `org.trellisldp.app.AbstractTrellisApplication<T extends TrellisConfiguration>` and write implementations for:

```java
ResourceService getResourceService();

IOService getIOService();

BinaryService getBinaryService();

Optional<AuditService> getAuditService();

Optional<BinaryService.MultipartCapable> getMultipartUploadService();
```

Returning `Optional.empty` for the last two methods is completely legit. There is also this method that can be overridden, if desired:

```java
void initialize(T config, Environment environment);
```

@ajs6f will this work for your impl?


acoburn:
Yes, it's mostly a lot of moving things around. It means that the `trellis-app` module doesn't generate a deployable artifact any longer; rather the `trellis-app-triplestore` does all of that.

ajs6f:
One suggestion-- a Javadoc on `::initialize` explaining when it gets called might be nice, since it has no annotation or overridden method governing that.

ajs6f:
That's good-- no need for other deployable apps to pull all of `-triplestores` dependencies.

ajs6f:
LGTM!

christopher-johnson:
small thing: It is not necessary to declare the test dependencies:
https://github.com/trellis-ldp/trellis/blob/d56b3fd60dad1d5c67c5ebb557191c547093242d/trellis-app/build.gradle#L28-L29
since they are in the root project build.gradle

acoburn:
@christopher-johnson thanks for catching that!

christopher-johnson:
also, a very minor detail is the syntax of those declarations:
https://github.com/trellis-ldp/trellis/blob/master/build.gradle#L116-L118
can be changed to this (without the closure "{..}"):
```groovy
        testImplementation("org.junit.jupiter:junit-jupiter-api:$junitVersion")
        testImplementation("org.apiguardian:apiguardian-api:$apiguardianVersion")
        testRuntime("org.junit.jupiter:junit-jupiter-engine:$junitVersion")
```
this is new since the latest gradle version I guess.  It looks better ...

acoburn:
@christopher-johnson let's put the gradle syntax changes in a separate ticket, and feel free to make a PR for normalizing how dependencies are defined. At present it's not particularly consistent.

christopher-johnson:
OK.  quick question: do you prefer the verbose dependency syntax or the abbreviated one?  I like the abbreviated one because it is easy to copy to modules.  The verbose syntax is also a bit annoying to enter, unless one copies it from https://mvnrepository.com/artifact/org.trellisldp/ ...

acoburn:
Re: the verbose/compact dependency syntax, I really don't have much of an opinion. I think the compact syntax is more common, and if you prefer that, feel free to change everything over to the compact form

PR #113, opened 2018-04-19
Use compact dependency syntax for Gradle

### LINKED ISSUE #112: Use compact dependency syntax for Gradle
The syntax used for defining dependencies in Gradle is not consistent, and there is no particular reason that the verbose syntax is being used. It would be better to make use of the compact syntax throughout the various `build.gradle` files.

christopher-johnson:
note there are a few optional jdk8 (but required in jpms) additions.  this is not a complete list of jpms required additions, just a step towards consolidation.  I added `force true` for some of the key transitive dependencies.  (for testing, it probably makes sense).   I also added a TODO regarding `jena-osgi`. 
 
there should be no subtractions, so everything should work the same.  (please verify this ..)  
closes #112


christopher-johnson:
the build is failing because of the OSGITest.  Odd, because the only difference in the build.gradle is that I changed from TestCompile to TestImplementation.  Perhaps, the features thing does not work with the current gradle DSL.  I will change it back and retest it.

trellis-app-triplestore/build.gradle
```diff
@@ -8,48 +8,50 @@ mainClassName = 'org.trellisldp.app.triplestore.TrellisApplication'
...
-    testCompile group: 'io.dropwizard', name: 'dropwizard-client', version: dropwizardVersion
-    testCompile group: 'io.dropwizard', name: 'dropwizard-testing', version: dropwizardVersion
-    testCompile group: 'org.apache.activemq', name: 'activemq-broker', version: activeMqVersion
-    testCompile group: 'org.apiguardian', name: 'apiguardian-api', version: apiguardianVersion
-    testCompile group: 'org.apache.tamaya', name: 'tamaya-core', version: tamayaVersion
-    testCompile group: 'org.awaitility', name: 'awaitility', version: awaitilityVersion
-    testCompile group: 'ch.qos.logback', name: 'logback-classic', version: logbackVersion
-    testCompile group: 'org.mockito', name: 'mockito-core', version: mockitoVersion
-    testCompile project(':trellis-test')
+    implementation("com.github.jsonld-java:jsonld-java:$jsonldVersion") {force = true}
```
acoburn: Remove the `{force = true}`

trellis-app-triplestore/build.gradle
```diff
@@ -8,48 +8,50 @@ mainClassName = 'org.trellisldp.app.triplestore.TrellisApplication'
...
+    implementation project(':trellis-file')
+    implementation project(':trellis-namespaces')
+    implementation project(':trellis-agent')
+    implementation project(':trellis-audit')
+    implementation project(':trellis-event-serialization')
+    implementation project(':trellis-webac')
+    implementation project(':trellis-triplestore')
+    implementation project(':trellis-jms')
+    implementation project(':trellis-kafka')
+    implementation project(':trellis-app')
```
acoburn: Because this is a deployable application (rather than a library), I'd prefer to use the `compile` scope for the above dependencies.

trellis-app-triplestore/build.gradle
```diff
@@ -8,48 +8,50 @@ mainClassName = 'org.trellisldp.app.triplestore.TrellisApplication'
...
+
+    testImplementation("com.github.jsonld-java:jsonld-java:$jsonldVersion") {force = true}
+    testImplementation("io.dropwizard:dropwizard-client:$dropwizardVersion")
+    testImplementation("io.dropwizard:dropwizard-testing:$dropwizardVersion")
+    testImplementation("org.apache.activemq:activemq-broker:$activeMqVersion")
+    testImplementation("org.apache.tamaya:tamaya-core:$tamayaVersion")
+    testImplementation("org.awaitility:awaitility:$awaitilityVersion")
+    testImplementation("ch.qos.logback:logback-classic:$logbackVersion")
+    testImplementation("org.mockito:mockito-core:$mockitoVersion")
+    testImplementation project(':trellis-test')
```
acoburn: `testCompile` as above

trellis-app/build.gradle
```diff
@@ -5,29 +5,31 @@ ext {
...
-    testImplementation group: 'io.dropwizard', name: 'dropwizard-client', version: dropwizardVersion
-    testImplementation group: 'io.dropwizard', name: 'dropwizard-testing', version: dropwizardVersion
-    testImplementation group: 'org.apache.commons', name: 'commons-rdf-jena', version: commonsRdfVersion
-    testImplementation group: 'org.apache.jena', name: 'jena-osgi', version: jenaVersion
-    testImplementation group: 'org.apache.jena', name: 'jena-tdb2', version: jenaVersion
-    testImplementation group: 'org.awaitility', name: 'awaitility', version: awaitilityVersion
-    testImplementation group: 'org.bouncycastle', name: 'bcprov-jdk15on', version: bouncycastleVersion
-    testImplementation group: 'org.mockito', name: 'mockito-core', version: mockitoVersion
+    testImplementation("ch.qos.logback:logback-classic:$logbackVersion")
+    testImplementation ("commons-codec:commons-codec:$commonsCodecVersion") {force = true}
```
acoburn: Is commons-codec this really a dependency here?

trellis-file/build.gradle
```diff
@@ -7,22 +7,26 @@ ext {
...
+    implementation("org.apache.tamaya:tamaya-api:$tamayaVersion")
+    implementation("org.slf4j:slf4j-api:$slf4jVersion")
     implementation project(':trellis-vocabulary')
 
-    testImplementation group: 'ch.qos.logback', name: 'logback-classic', version: logbackVersion
-    testImplementation group: 'org.apache.tamaya', name: 'tamaya-core', version: tamayaVersion
-    testImplementation group: 'org.mockito', name: 'mockito-core', version: mockitoVersion
+    testImplementation("ch.qos.logback:logback-classic:$logbackVersion")
+    //Note JSON-LD is transitive dependency, but the version can be managed here
+    testImplementation("com.github.jsonld-java:jsonld-java:$jsonldVersion") {force = true}
```
acoburn: remove `{force = true}`

trellis-http/build.gradle
```diff
@@ -7,39 +7,39 @@ ext {
...
+    api("org.glassfish.hk2.external:javax.inject:$javaxInjectVersion")
     api project(':trellis-api')
 
-    implementation group: 'commons-codec', name: 'commons-codec', version: commonsCodecVersion
-    implementation group: 'commons-io', name: 'commons-io', version: commonsIoVersion
-    implementation group: 'javax.json', name: 'javax.json-api', version: javaxJsonVersion
-    implementation group: 'org.apache.tamaya', name: 'tamaya-api', version: tamayaVersion
-    implementation group: 'org.slf4j', name: 'slf4j-api', version: slf4jVersion
+    implementation ("commons-codec:commons-codec:$commonsCodecVersion") {force = true}
+    implementation ("commons-io:commons-io:$commonsIoVersion") {force = true}
```
acoburn: ditto

trellis-http/build.gradle
```diff
@@ -7,39 +7,39 @@ ext {
...
-    testImplementation group: 'com.google.guava', name: 'guava', version: guavaVersion
-    testImplementation group: 'org.apache.commons', name: 'commons-text', version: commonsTextVersion
-    testImplementation group: 'org.apache.commons', name: 'commons-rdf-simple', version: commonsRdfVersion
-    testImplementation group: 'org.apache.tamaya', name: 'tamaya-core', version: tamayaVersion
-    testImplementation group: 'org.glassfish', name: 'javax.json', version: jsonVersion
-    testImplementation group: 'org.glassfish.jersey.core', name: 'jersey-server', version: jerseyVersion
-    testImplementation group: 'org.glassfish.jersey.test-framework.providers', name: 'jersey-test-framework-provider-inmemory', version: jerseyVersion
-    testImplementation group: 'org.mockito', name: 'mockito-core', version: mockitoVersion
+    testImplementation("com.fasterxml.jackson.core:jackson-core:$jacksonVersion") {force = true}
+    testImplementation("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion") {force = true}
```
acoburn: ditto

trellis-io-jena/build.gradle
```diff
@@ -7,20 +7,22 @@ ext {
 }
 
 dependencies {
-    api group: 'org.apache.commons', name: 'commons-rdf-api', version: commonsRdfVersion
-    api group: 'org.glassfish.hk2.external', name: 'javax.inject', version: javaxInjectVersion
+    api("org.apache.commons:commons-rdf-api:$commonsRdfVersion")
+    api("org.glassfish.hk2.external:javax.inject:$javaxInjectVersion")
+    api("commons-io:commons-io:$commonsIoVersion") {force = true}
```
acoburn: remove `{force = true}`

acoburn: `commons-io` is an implementation dependency

trellis-namespaces/build.gradle
```diff
@@ -8,14 +8,15 @@ ext {
 
 dependencies {
     api project(':trellis-api')
-    api group: 'org.glassfish.hk2.external', name: 'javax.inject', version: javaxInjectVersion
+    api("org.glassfish.hk2.external:javax.inject:$javaxInjectVersion")
 
-    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: jacksonVersion
-    implementation group: 'org.slf4j', name: 'slf4j-api', version: slf4jVersion
-    implementation group: 'org.apache.tamaya', name: 'tamaya-api', version: tamayaVersion
+    implementation ("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion") {force = true}
```
acoburn: remove `{force = true}`

trellis-app-triplestore/build.gradle
```diff
@@ -8,48 +8,50 @@ mainClassName = 'org.trellisldp.app.triplestore.TrellisApplication'
...
+    implementation project(':trellis-event-serialization')
+    implementation project(':trellis-webac')
+    implementation project(':trellis-triplestore')
+    implementation project(':trellis-jms')
+    implementation project(':trellis-kafka')
+    implementation project(':trellis-app')
+
+    runtime ("javax.xml.bind:jaxb-api:$jaxbVersion")
+
+    testImplementation("com.github.jsonld-java:jsonld-java:$jsonldVersion") {force = true}
```
acoburn: This is already included above

trellis-app/build.gradle
acoburn: remove `{force = true}`

trellis-event-serialization/build.gradle
```diff
@@ -7,16 +7,16 @@ ext {
...
-    api group: 'org.apache.commons', name: 'commons-rdf-api', version: commonsRdfVersion
+    api("org.apache.commons:commons-rdf-api:$commonsRdfVersion")
     api project(':trellis-api')
 
-    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-core', version: jacksonVersion
-    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: jacksonVersion
-    implementation group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-jsr310', version: jacksonVersion
+    implementation("com.fasterxml.jackson.core:jackson-core:$jacksonVersion") {force = true}
+    implementation("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion") {force = true}
+    implementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jacksonVersion") {force = true}
```
acoburn: remove `{force = true}`

trellis-app-triplestore/build.gradle
christopher-johnson: I think that compile and testCompile are deprecated.  the only current options are api and implementation.  Check the docs here: https://docs.gradle.org/current/userguide/java_library_plugin.html

acoburn: That’s for the java-library plugin. This should be using the java plugin

trellis-app/build.gradle
christopher-johnson: If i remember correctly, I think that `testPostBinaryWithDigest` asks for a digest.  Anything that invokes the MessageDigest library will require the codec.  In this case, since it is a runtime classpath situation with this test, the app would require it (in an encapsulated environment) unless trellis-file opened codec to the app.

acoburn: I believe those tests have all been moved to `trellis-test` and so this dependency is not needed here.

trellis-app-triplestore/build.gradle
acoburn: At present, the root build script applies the `java` and `java-library` plugin to _all_ subprojects. It would probably be better to selectively apply those plugins to the individual projects, as is currently done with the `osgi` plugin.

christopher-johnson: I removed the java plugin from the root build script and applied it only to `trellis-app-triplestore` and `trellis-osgi`.  These are the only modules that require `compile`, correct?  I understand your motivation here.

christopher-johnson:
I apologize for the branch confusion, which I guess is why appveyor failed.  I have different remotes configured at home and at work... somewhat difficult to remember where I am sometimes.

acoburn:
This looks good. One more suggestion: I would invert what you did with the `java` and `java-library` plugins. That is, leave the `java` plugin in the root Gradle configuration and apply the `java-library` plugin in each relevant project (if you don't have time for this, I can also just merge this as-is and make this change as a follow-on). The reasoning is that the `java-library` plugin is an extension of the `java` plugin so it would make sense to put the most generic one in the root configuration and have the more specialized one in the relevant projects (i.e. all but -app-triplestore and -osgi).

christopher-johnson:
I wonder if it might make sense to remove trellis-app-triplestore from the multiproject library build and assemble it separately from release or snapshot artifacts.  This would speed things up.  It may also make sense to have a sonatype mechanism for publishing snapshots.  This interests me since a client library could pull the snapshots rather than having to load them locally.

christopher-johnson:
Actually, I see that there is the `snapshotRepository` configuration in the `uploadArchives` task already.  Is it possible to distinguish snapshots by branch?  For example, I would like to publish the JPMS branch 0.7.0-SNAPSHOT libraries for the client.

christopher-johnson:
I guess that this could be done easily with the version, e.g. {$branchName}-SNAPSHOT in gradle.properties.  It does not make sense to version branch snapshots either, what is needed is a timestamp, that would be available in the sonatype artifact.  This would also help with docker deployment.

christopher-johnson:
this can be done by adding
```groovy
 appendix 'jpms'
```
to the jar configuration.

acoburn:
I would not advocate for splitting the trellis-app-triplestore module out into a separate project. The trellis codebase used to be split out across lots of little projects, and it just confused anyone who wanted to contribute. It is also confusing if the project doesn’t create any deployable artifacts. Finally, the tests that are run in that module are really important for confirming the correctness of the other libraries: it is where the end-to-end integration tests are run.

Snapshot artifacts are already pushed to sonatype. I have to do this manually, though, and so it doesn’t happen with every build. That’s fine if you’d like to publish jpms snapshots (so long as the artifact ids are distinct from the rest of the artifacts).

ajs6f:
For the future, we might want to bless a memory-resident backend as a trusted test framework. Then we could do the tests without a full impl. OTOH, having a full real-world impl do the tests has its own value.

christopher-johnson:
ok on the jpms snapshots.  I would like to do this soon.  How about I push one jpms snapshot artifact and you can review it to make sure it is ok?

The `application` task is very slow.  Maybe there is something more to do to improve the build performance, like incremental compilation.  Alternatively, the module could stay in the multiproject build for testing, but the distribution and application packaging could be separated.

acoburn:
It would be helpful to quantify what you mean by "slow". Assuming that the dependencies have been cached locally, calling `gradle assemble` on the trellis-app-triplestore project takes 15-17 seconds on my laptop. It is less if you exclude the `javadoc` task (`gradle assemble -x javadoc`). Running all of the tests (`gradle check assemble`) brings that total to over a minute, so for me, most of the time is taken up with the unit tests, not the application packaging. Whether that constitutes "slow" is, I suppose, up for interpretation.

That said, moving the application/distribution configuration to a separate module is a very good idea.

christopher-johnson:
I was assuming that nothing was cached.  The build time is not consistent so it is difficult to quantify.  The JPMS builds seem to be faster by about 10 mins than master, not sure why.  You are right in that the assembly task is fast, but to me, it seems like the javadoc tasks take forever when they do execute.

acoburn:
Judging by the builds on Travis-CI -- where all dependencies are downloaded each time, the builds on master take about 8-10 minutes. A complete build, locally, (with all dependencies cached and running the full test suite) takes about half of that (4-5 minutes). OTOH, simply compiling all of the artifacts (without testing: `gradle assemble`) takes about 60 seconds for me and omitting the javadoc task cuts that in half (~30 seconds).

christopher-johnson:
https://travis-ci.org/trellis-ldp/trellis/builds says the master build takes 29 minutes

acoburn:
Right: it's between 8 and 10 minutes for each of jdk8, jdk9 and jdk10, for a combined total of ~30 minutes (jdk8 is longer b/c it also handles the sonarqube and coveralls tasks), but the elapsed clock time is more like 10 minutes b/c the jobs are run in parallel. In that context, the jpms branch takes nearly twice as long to build. See, for example, this recent build from master: https://travis-ci.org/trellis-ldp/trellis/builds/368708399

christopher-johnson:
oh, I guess my sense of time is distorted.  I am very impatient ...  

On a different subject, I think I have found an SSL related problem with the maven plugin, but perhaps you can confirm.  I am able to upload to sonatype using JDK8, but when I use JDK10, I get the dreaded `pkix path building failed`.  Yes, I have added the sonatype/godaddy certificate chain to the keystore.  I have also tried to override the Apache Wagon ssl properties and that did not work either.  With the Apache Client (which is what Apache Wagon is using), I can GET from Sonatype only if I set t up like this:
```java
final SSLContext sslContext = new SSLContextBuilder()
      .loadTrustMaterial(null, (x509CertChain, authType) -> true).build();
CloseableHttpClient httpClient = HttpClients.custom()
      .setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE)
      .setSSLContext(sslContext).build();
```
Do know if it is possible to configure the Maven plugin with an sslContext for Sonatype?

acoburn:
I have not had any issues with SSL certs while uploading to Sonatype w/ JDK10. When cutting releases I typically use JDK8, but these days I typically publish snapshot artifacts w/ JDK10. I have found that the root cacert for a linux-installed JDK10 includes fewer trusted certificates than what comes with the root cacert on an OSX-installed JDK10.

christopher-johnson:
That is good for you using OSX.   But the cacert JKS is directly supplied in the JDK no?  I downloaded it directly from Oracle, so I do not know why I would have this trouble.

I have done some digging into how the MavenDeployAction actually works in Gradle.  You might be surprised to know that the [org.sonatype.aether dependency](https://github.com/gradle/gradle/blob/master/gradle/dependencies.gradle#L25)  is from 2011.  I tried to use the documented maven plugin deployment artifact configuration and replaced the http wagon with one that I built, but it does not work... I am thinking that this is a gradle problem.

acoburn:
I don't believe the `gradle upload` task has anything to do with that sonatype plugin. Trellis uses the `maven-publish` plugin, and though it is pushing artifacts to `https://oss.sonatype.org`, it could be pushing them anywhere.

I would use the `keytool` application to check the root cacert to find out if there is a certificate for GoDaddy: `keytool -list -keystore /path/to/cacert` If not, you should add it.

christopher-johnson:
I attach the stacktrace for reference.  [upload-exception.txt](https://github.com/trellis-ldp/trellis/files/1933101/upload-exception.txt)  I definitely have the cert installed, though it could be an issue with the GoDaddy cert itself.  One way to confirm this would be to test gradle to a different https nexus.

From what I can decipher, the maven-publish plugin generates the pom.xml but it seems gradle delegates the deployment to the upload task which is handled by `sonatype.aether`. 

What command do you use to deploy?  I just used `./gradlew --stacktrace trellis-webac:upload`

acoburn:
I use `gradle upload`.

The stacktrace indicates that your `cacert` does not contain a certificate for `oss.sonatype.org`. You may want to verify that your JVM is, in fact, reading from the cacert that you think it's reading from. And if so, verify that the correct cert chains are present.

christopher-johnson:
update on this:
I have determined that this is certificate related.  I made a test that uses bouncycastle here https://github.com/christopher-johnson/x509-validator
It is not complete yet, as I will add a method to create a fresh jks rather than the local one.  Of the test certs in resources, the only one that passes is the amazon root.  I still need to figure out how the intermediate chain is supposed to work, but my guess is that there is some magic to this chain that the sonatype cert is missing.  The cert does include a chain though, where the other samples do not.  This cert provider and protocol is an enigma and it is generally undocumented.  I will decipher it eventually, though I would rather not have to... Sonatype and any https server should just provide something that works without the client having to become an encryption guru.

christopher-johnson:
Also, i can add here that I need to defer the jpms snapshot push anyway because there are more package problems with jena.  The refactoring required to support modules in jena will be substantial.  I am working on a fork that will use gradle as upstream will probably not be able to do this for awhile.

christopher-johnson:
Regarding this PR, is there something else you need here?

acoburn:
@christopher-johnson see [this comment](https://github.com/trellis-ldp/trellis/pull/113#issuecomment-382918375): I would keep the `java` plugin in the root build config and move the `java-library` plugin to the various subprojects (since `java-library` extends `java`)

christopher-johnson:
Ok.  Although I thought that gradle works by configuration injection.  Meaning the allprojects configuration is injected into the subprojects (as required).  So, if the java plugin is already there in the subproject and there are no library related configs the build task moves on.  Also, the test dependencies in the root configuration use the java library config.

Maybe this can be done later with a separate PR/ commit that creates a module for distribution?

Issue #91, opened 2018-04-03
Who creates containment triples?


ajs6f:
Having enabled the new `ResourceServiceTests` I'm getting a failure at https://github.com/trellis-ldp/trellis/blob/651bd25595cb2bccee42a1f82e7b9845712c9e65/trellis-test/src/main/java/org/trellisldp/test/ResourceServiceTests.java#L410
of the form `expected: <2> but was: <0>`. Presumably this is because I am not producing the containment triples from the backend (which my backend is not, in fact, doing). But I was under the impression that it is `trellis-http` that is responsible for creating the containment triples and passing them into the backend. WRT quads, my backend just serializes what it is given and deserializes it (with the exceptional case of an LDP-NR, when it parses the quads to get some metadata).

Am I missing something? How can the `ResourceServiceTests` expect to find containment triples in the absence of `trellis-http`?

acoburn:
The HTTP layer does not generate the containment quads. There are several reasons for this related to both concurrency and scale. In the triplestore impl, I store a child -> parent link on each resource (rather than the LDP parent -> child triple), and then dynamically generate the containment triples by running a query that reverses that relationship. If it would be helpful, we could certainly have the HTTP layer add this child -> parent link. But having the http layer keep track of all of a resource's children would be rather problematic.

ajs6f:
Ok, I see that. But then how does the backend know how to produce those links? The backend may have no idea what a resource's container is. And supposing it does know, is it responsible for adding (or removing) membership triples to the parent?

acoburn:
That's the role of the `ResourceService::getContainer` method. A backend can override the default impl if it has a different way to keep track of containment.

ajs6f:
I can see that `ResourceService::getContainer` does fine to _return_ the containment info out of the backend. But how does it get there to begin with?

acoburn:
One of the distinct advantages to having you independently implement a persistence layer is that you bring to the fore all of my built-in assumptions about the inner working of this code. This is another such case where the persistence layer is simply assuming that containment works in a particular way. As it is, there is nothing -- _absolutely nothing_ -- in the code to inform a newly created child resource of the existence of its parent. I suspect that the best way to address this would be to have the HTTP layer explicitly add that `trellis:PreferServerManaged <child> dc:isPartOf <parent> .` quad to the dataset that is passed along to `::create`, `::delete` and `::replace`.

ajs6f:
[...]

PR #120, opened 2018-05-01
Adding container parameter to mutating code paths

ajs6f:
Resolves https://github.com/trellis-ldp/trellis/issues/91

### DIFF
--- a/components/test/src/main/java/org/trellisldp/test/ResourceServiceTests.java
+++ b/components/test/src/main/java/org/trellisldp/test/ResourceServiceTests.java
@@ -88,7 +88,7 @@ public interface ResourceServiceTests {
         dataset.add(Trellis.PreferUserManaged, identifier, type, SKOS.Concept);
 
         assertFalse(getResourceService().get(identifier).isPresent());
-        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, dataset).get());
+        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, null, dataset).get());
         final Optional<? extends Resource> res = getResourceService().get(identifier);
         assertTrue(res.isPresent());
         res.ifPresent(r -> r.stream(Trellis.PreferUserManaged)
@@ -110,14 +110,14 @@ public interface ResourceServiceTests {
         dataset.add(Trellis.PreferUserManaged, identifier, type, SKOS.Concept);
 
         assertFalse(getResourceService().get(identifier).isPresent());
-        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, dataset).get());
+        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, null, dataset).get());
 
         dataset.clear();
         dataset.add(Trellis.PreferUserManaged, identifier, SKOS.prefLabel, rdf.createLiteral("preferred label"));
         dataset.add(Trellis.PreferUserManaged, identifier, SKOS.altLabel, rdf.createLiteral("alternate label"));
         dataset.add(Trellis.PreferUserManaged, identifier, type, SKOS.Concept);
 
-        assertTrue(getResourceService().replace(identifier, getSession(), LDP.RDFSource, dataset).get());
+        assertTrue(getResourceService().replace(identifier, getSession(), LDP.RDFSource, null, dataset).get());
         final Optional<? extends Resource> res = getResourceService().get(identifier);
         assertTrue(res.isPresent());
         res.ifPresent(r -> {
@@ -143,7 +143,7 @@ public interface ResourceServiceTests {
 
         assertFalse(getResourceService().get(identifier).isPresent());
 
-        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, dataset).get());
+        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, null, dataset).get());
         assertTrue(getResourceService().get(identifier).filter(res -> !res.isDeleted()).isPresent());
 
         assertTrue(getResourceService().delete(identifier, getSession(), LDP.Resource, rdf.createDataset()).get());
@@ -162,7 +162,7 @@ public interface ResourceServiceTests {
         final Dataset dataset0 = rdf.createDataset();
         dataset0.add(Trellis.PreferUserManaged, identifier, DC.title, rdf.createLiteral("Immutable Resource Test"));
 
-        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, dataset0).get());
+        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, null, dataset0).get());
 
         final IRI audit1 = rdf.createIRI(TRELLIS_BNODE_PREFIX + getResourceService().generateIdentifier());
         final Dataset dataset1 = rdf.createDataset();
@@ -221,7 +221,7 @@ public interface ResourceServiceTests {
         dataset.add(Trellis.PreferUserManaged, identifier, DC.subject, rdf.createIRI(SUBJECT1));
 
         assertFalse(getResourceService().get(identifier).isPresent());
-        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, dataset).get());
+        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, null, dataset).get());
         final Optional<? extends Resource> res = getResourceService().get(identifier);
         assertTrue(res.isPresent());
         res.ifPresent(r -> {
@@ -267,7 +267,7 @@ public interface ResourceServiceTests {
         dataset.add(Trellis.PreferServerManaged, binaryLocation, DC.extent, rdf.createLiteral("150", XSD.long_));
 
         assertFalse(getResourceService().get(identifier).isPresent());
-        assertTrue(getResourceService().create(identifier, getSession(), LDP.NonRDFSource, dataset).get());
+        assertTrue(getResourceService().create(identifier, getSession(), LDP.NonRDFSource, null, dataset).get());
         final Optional<? extends Resource> res = getResourceService().get(identifier);
         assertTrue(res.isPresent());
         res.ifPresent(r -> {
@@ -313,7 +313,7 @@ public interface ResourceServiceTests {
         dataset0.add(Trellis.PreferUserManaged, identifier, DC.subject, rdf.createIRI(SUBJECT0));
 
         assertFalse(getResourceService().get(identifier).isPresent());
-        assertTrue(getResourceService().create(identifier, getSession(), LDP.Container, dataset0).get());
+        assertTrue(getResourceService().create(identifier, getSession(), LDP.Container, null, dataset0).get());
 
         final IRI child1 = rdf.createIRI(base + "/child01");
         final Dataset dataset1 = rdf.createDataset();
@@ -321,7 +321,7 @@ public interface ResourceServiceTests {
         dataset1.add(Trellis.PreferUserManaged, child1, DC.subject, rdf.createIRI(SUBJECT1));
 
         assertFalse(getResourceService().get(child1).isPresent());
-        assertTrue(getResourceService().create(child1, getSession(), LDP.RDFSource, dataset1).get());
+        assertTrue(getResourceService().create(child1, getSession(), LDP.RDFSource, null, dataset1).get());
 
         final IRI child2 = rdf.createIRI(base + "/child02");
         final Dataset dataset2 = rdf.createDataset();
@@ -329,7 +329,7 @@ public interface ResourceServiceTests {
         dataset2.add(Trellis.PreferUserManaged, child2, DC.subject, rdf.createIRI(SUBJECT2));
 
         assertFalse(getResourceService().get(child2).isPresent());
-        assertTrue(getResourceService().create(child2, getSession(), LDP.RDFSource, dataset2).get());
+        assertTrue(getResourceService().create(child2, getSession(), LDP.RDFSource, null, dataset2).get());
 
         final Optional<? extends Resource> res = getResourceService().get(identifier);
         assertTrue(res.isPresent());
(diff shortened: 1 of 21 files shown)


core/api/src/main/java/org/trellisldp/api/ImmutableDataService.java
```diff
@@ -28,11 +29,13 @@
...
     /**
-     * @param identifier the identifier under which to persist a resource
+     * @param identifier the identifier under which to persist a dataset
      * @param session the session context for this operation
-     * @param resource a resource to persist
+     * @param dataset a dataset to persist
      * @return whether the resource was successfully persisted
      */
-    Future<Boolean> add(IRI identifier, Session session, T resource);
+    Future<Boolean> add(IRI identifier, Session session, Dataset dataset);
```
acoburn: Since the interface is no longer using the generic `<T>` type, could we remove that from the definition above?

core/api/src/main/java/org/trellisldp/api/MutableDataService.java
```diff
@@ -27,27 +28,39 @@
...
      * @param session the session context for this operation
-     * @param resource a resource to persist
-     * @return whether the resource was successfully persisted
+     * @param ixnModel the LDP interaction model for this resource
+     * @param container an LDP container for this resource
+     * @param dataset the dataset
+     * @return whether the resource was added
      */
-    Future<Boolean> create(IRI identifier, Session session, U resource);
+    Future<Boolean> create(IRI identifier, Session session, IRI ixnModel, IRI container, Dataset dataset);
```
acoburn: See above about removing the generic `<U>` type from the interface definition, since it appears that it is no longer used.

core/http/src/main/java/org/trellisldp/http/MultipartUploader.java
```diff
@@ -283,7 +283,8 @@ public Response createBinary(@PathParam("id") final String id, @Context final Co
                             upload.getBinary().getIdentifier(), DC.extent,
                             rdf.createLiteral(size.toString(), XSD.long_))));
 
-            if (resourceService.create(identifier, session, NonRDFSource, dataset.asDataset()).get()) {
+            final IRI container = resourceService.getContainer(rdf.createIRI(TRELLIS_DATA_PREFIX + id)).orElse(null);
+            if (resourceService.create(identifier, session, NonRDFSource, container  , dataset.asDataset()).get()) {
```
acoburn: minor: the `container    ,` might not play well with checkstyle.

ajs6f:
Please cast a narrow eye on the HTTP stuff. I have a lot of time keeping internal and external IDs straight, so I might have mushed things up in a way that passes tests but is still wrong.

core/http/src/main/java/org/trellisldp/http/impl/PostHandler.java
```diff
@@ -152,8 +153,10 @@ public ResponseBuilder createResource() {
                 // Check for any constraints
                 checkConstraint(dataset, PreferUserManaged, ldpType, rdfSyntax.orElse(TURTLE));
             }
-
-            if (resourceService.create(internalId, session, ldpType, dataset.asDataset()).get()) {
+            final IRI container = resourceService.getContainer(internalId).orElse(null);
```
acoburn: This is perfectly fine as it is, but it might be slightly more efficient (and clear, code-wise) to create the container from data we've already got in the `POST` operation. I.e.:

```java
final IRI container = rdf.createIRI(TRELLIS_DATA_PREFIX + req.getPath());
```

core/api/src/main/java/org/trellisldp/api/ImmutableDataService.java
ajs6f: Wait, no-- the type param is passed through to `RetrievalService<T>`. So we can either leave it here, or stop extending `RetrievalService<T>` here (and in the mutable one) and have other classes implement `RetrievalService<T>` independently. WDYT?

acoburn: In that case, I'd probably leave this as-is for this PR. I wouldn't mind looking more closely at our use of those generics later -- they can probably be simplified, but that's really getting into a separate issue.

core/api/src/main/java/org/trellisldp/api/MutableDataService.java
```diff
@@ -27,27 +28,39 @@
...
     /**
-     * @param identifier the identifier for the resource to persist
+     * Put a resource into the server.
+     *
+     * @param identifier the identifier for the new resource
      * @param session the session context for this operation
-     * @param resource a resource to persist
-     * @return whether the resource was successfully persisted
+     * @param ixnModel the LDP interaction model for this resource
+     * @param container an LDP container for this resource
```
acoburn: One more thing -- it would be really helpful if the javadoc here indicates that the container parameter can be `null`.

core/api/src/main/java/org/trellisldp/api/MutableDataService.java
```diff
@@ -27,27 +28,39 @@
...
     /**
-     * @param identifier the identifier for the resource to persist
+     * Replace a resource in the server.
+     *
+     * @param identifier the identifier for the new resource
      * @param session the session context for this operation
-     * @param resource a resource to persist
-     * @return whether the resource was successfully persisted
+     * @param ixnModel the LDP interaction model for this resource
+     * @param container an LDP container for this resource
```
acoburn: ditto with the container javadoc

core/http/src/main/java/org/trellisldp/http/MultipartUploader.java
acoburn: I amended the checkstyle config to catch these: 474df51a0661f0eb70c5d7904a93bbf9a1bfe72d

core/api/src/main/java/org/trellisldp/api/ImmutableDataService.java
ajs6f: Ok, cool. We can revisit, of course.

core/api/src/main/java/org/trellisldp/api/MutableDataService.java
ajs6f: Good catch. What about `javax.annotation.CheckForNull` or `javax.annotation.Nullable`? Worth adding?

components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -156,8 +156,8 @@ private static RDFTerm getBaseIRI(final RDFTerm object) {
     }
 
     @Override
-    public Future<Boolean> create(final IRI identifier, final Session session, final IRI ixnModel,
-            final Dataset dataset) {
+    public Future<Boolean> create(final IRI identifier, final Session session, final IRI ixnModel, final IRI container,
+                    final Dataset dataset) {
```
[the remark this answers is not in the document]
ajs6f: Done.

components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -175,8 +175,8 @@ private static RDFTerm getBaseIRI(final RDFTerm object) {
     }
 
     @Override
-    public Future<Boolean> replace(final IRI identifier, final Session session, final IRI ixnModel,
-            final Dataset dataset) {
+    public Future<Boolean> replace(final IRI identifier, final Session session, final IRI ixnModel, final IRI container,
+                    final Dataset dataset) {
```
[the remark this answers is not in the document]
ajs6f: Done.

core/api/src/main/java/org/trellisldp/api/MutableDataService.java
acoburn: I haven't been adding those annotations to other parts of the code, but it seems like a good idea for the API. Feel free to start here, and the relevant annotations can be added elsewhere by the by.

ajs6f: Hm, never mind, I have to walk it back. Those annotations aren't in our classpath, or at least not in such a way as to let me deploy them here. Oh, well, Javadocs are good too.

acoburn: Just a note that these annotations `javax.annotation.Nullable` are not part of the standard `annotation-api` artifact. We'd need to add a dependency on 'com.google.code.findbugs/jsr305' (JSR 305 itself seems pretty dormant). Is it worth adding that dependency?

ajs6f: Right, see my note above (we just crossed messages). If it _is_ worth adding that, it seems to me not to belong in this PR, eh?

acoburn: correct -- that would be a separate PR.

components/test/src/main/java/org/trellisldp/test/ResourceServiceTests.java
```diff
@@ -88,7 +88,7 @@ default void testCreateResource() throws Exception {
         dataset.add(Trellis.PreferUserManaged, identifier, type, SKOS.Concept);
 
         assertFalse(getResourceService().get(identifier).isPresent());
-        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, dataset).get());
+        assertTrue(getResourceService().create(identifier, getSession(), LDP.RDFSource, null, dataset).get());
```
acoburn: This `null` value is now going wreak havoc with the tests. Perhaps not with this particular test, but it will generally -- you can see the CI failures as an indication of that. I'd suggest replacing it with either: `getResourceService().getContainer(identifier)` or `final IRI container = rdf.createIRI(TRELLIS_DATA_PREFIX);`

The later option you could add as a class-wide field that gets re-used.

acoburn: On second thought, the first option will be the most flexible. Just replace `null` with `getResourceService().getContainer(identifier).orElse(null)`

components/triplestore/src/test/java/org/trellisldp/triplestore/TriplestoreResourceServiceTest.java
```diff
@@ -237,7 +237,7 @@ public void testRDFConnectionError() throws Exception {
 
         final IRI resource = rdf.createIRI(TRELLIS_DATA_PREFIX + "resource");
         assertThrows(ExecutionException.class, () ->
-                svc.create(resource, mockSession, LDP.RDFSource, rdf.createDataset()).get());
+                svc.create(resource, mockSession, LDP.RDFSource, null, rdf.createDataset()).get());
```
acoburn: You'll find the same issue with tests in this class. Here, feel free to replace `null` with `rdf.createIRI(TRELLIS_DATA_PREFIX)`, which you can put into a private field.

components/test/src/main/java/org/trellisldp/test/ResourceServiceTests.java
ajs6f: Wait, what? All tests pass for me. I believe you, I just want to understand why this is of concern; the only thing I understand the `null` to mean is that this isn't a contained resource-- to my thinking, that shouldn't fail any tests, unless they are specifically checking for containment. (Also, WTF do they pass for me?)

acoburn: I have no idea why the tests pass for you. They fail for me locally.

Specifically, in the `org.trellisldp.test.ResourceServiceTests` there are four tests: one for each of LDP-C, LDP-BC, LDP-DC and LDP-IC where containment triples are explicitly being counted. You can see an example [here](https://github.com/trellis-ldp/trellis/blob/master/components/test/src/main/java/org/trellisldp/test/ResourceServiceTests.java#L411).

In the `org.trellisldp.triplestore.TriplestoreResourceServiceTest` class, there is an expectation (internally) that an identified container will result in an extra `PreferServerManaged` triple. The fix here is either to leave the `null` value in the method and decrement the triple count in the test or leave the triple count the same and replace the `null` with a real `IRI`. In this [example](https://github.com/trellis-ldp/trellis/blob/ContainerParam/components/triplestore/src/test/java/org/trellisldp/triplestore/TriplestoreResourceServiceTest.java#L312-L325), the test is expecting 3 server-managed triples and a total of 5 triples. Keeping the `null` value in place, you'd want to change line 321 to `2L` and line 324 to `4L`.

Which is to say: the `null` value is just fine so long as the tests are not expecting to count the containment-related triples.

ajs6f: Okay, that makes more sense-- counting containment triples is exactly what I _would_ expect to fail. But won't using `getResourceService().getContainer(identifier).orElse(null)` require more mocking (of that method in the `mockResourceService`)? Could I not just as well use a dummy IRI?

ajs6f: e.g. `info:testContainer` or the like

acoburn: In this case, you shouldn't need any mocks. This test code is set up so that a _working_ `ResourceService` is used, and so while `info:testContainer` would probably work, it might be better for the actual `ResourceService` instance to manage its own parent-container definitions.

ajs6f: Okay, I am now getting failures, so that's good (if really annoying). I will take a crack at the `getResourceService().getContainer(identifier).orElse(null)` approach.

ajs6f: Nope, didn't fix the fails. Come to think of it, here's a common declaration for the id of resources in the tests:
```
 final IRI resource = rdf.createIRI(TRELLIS_DATA_PREFIX + "resource")
```
Now when stored via the java API, am I wrong in thinking that such a resource isn't going to get a "container" around it because it hasn't got one according to the path-munging algo? IOW, there is no path segment to interpret as a container.

acoburn:
This now builds and the tests complete locally. I'll merge this as soon as the CI-tests pass, just in case I missed something.

ajs6f:
Good to squash and merge?

ajs6f:
Ah, cool. I'll leave it to you. Thanks for the save at the end!

acoburn:
Thank _you_ for all the work!

PR #122, opened 2018-05-01
Fix the rest of the tests

acoburn:
@ajs6f this should fix the rest of the tests.


ajs6f:
Whoa, rad! I didn't realize there was a `root` floating around.

PR #124, opened 2018-05-02
Vocabulary

christopher-johnson:
This adds terms and changes the namespace variable name from `URI` to `NS` (to avoid collision with `DC.URI`).


acoburn:
While NS will avoid collision with dc:URI, I don’t think it’s very future-proof. What about using a ::getNamespace method instead.

Also, it is by design that not every vocabulary term is included here. A lot of vocabularies have obsolete terms, and I’d rather not include those in code. And given the way this PR is structured, it is hard to tell what has been added. I would suggest first changing the URI field and then, separately, adding new terms.

acoburn:
Plus, this sort of change affects other modules. I’d suggest running the full test suite

christopher-johnson:
Good idea with ::getNamespace.  So "NS" would be replaced with getNamespace() in each class and there would be a single method like this?
```java
    /* Namespace */
    public static String getNamespace() {
        return "https://www.w3.org/ns/activitystreams#";
    }
```
I did not add the deprecated terms in VCARD.

christopher-johnson:
Should the namespace be a private field in the class?  I did not expect `LDP.URI` to be used.

christopher-johnson:
I close this PR and resubmit, splitting out the getNamespace from the term changes.

PR #125, opened 2018-05-02
adds static getNamespace method to vocabulary classes

christopher-johnson:
as discussed in #124.


core/vocabulary/src/main/java/org/trellisldp/vocabulary/ACL.java
```diff
@@ -25,31 +25,38 @@
  * @author acoburn
  */
 public final class ACL {
+    private static final String NS = "http://www.w3.org/ns/auth/acl#";
```
acoburn: If you change `NS` for `URI` (effectively, making `URI` a private field), you'll make this PR much smaller. Clearly, for the `DC` class, you'd need a value such as `NS`, but for most classes the `URI` field won't conflict.

core/vocabulary/src/test/java/org/trellisldp/vocabulary/AbstractVocabularyTest.java
```diff
@@ -99,16 +99,13 @@ public void testVocabularyRev() {
...
-        assertEquals(namespace(), uri.get().get(null), "Namespaces do not match!");
+        final Method m = vocabulary().getMethod("getNamespace");
+        assertEquals(namespace(), m.invoke(null), "Namespaces do not match!");
     }
 
     private Stream<String> fields() {
         return stream(vocabulary().getFields()).map(Field::getName).map(name ->
                 name.endsWith("_") ? name.substring(0, name.length() - 1) : name)
-            .map(name -> name.replaceAll("_", "-")).filter(field -> !field.equals("URI"));
+            .map(name -> name.replaceAll("_", "-")).filter(field -> !field.equals("NS"));
```
acoburn: The `Class::getFields` method will return all of the public fields of the class, so there is no more need to filter by `field -> !field.equals("NS")`; in fact, that filter should be entirely removed, since "NS" may be used at some point as a public field.

acoburn:
Thanks @christopher-johnson

PR #127, opened 2018-05-02
Simplify Binary metadata handling in HTTP layer

### LINKED ISSUE #126: Simplify handling of Binary objects in the API
At present, metadata for `Binary` objects (location, size, mime-type, etc) are explicitly added to the internal `Dataset` in the `trellis:PreferServerManaged` graph. It would simplify the `ResourceService::create` and `ResourceService::replace` if one of the parameters is a (nullable) `Binary` object. Then, a persistence implementation can do whatever it needs to do with that data. It also gets the HTTP layer entirely out of the business of writing server-managed triples.

acoburn:
Resolves #126 

The changes in the test code are pretty uninteresting. Of note are the changes in `core/api` and `core/http`.

### DIFF
--- a/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
+++ b/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
@@ -90,6 +90,7 @@ import org.apache.jena.sparql.syntax.ElementPathBlock;
 import org.apache.jena.update.Update;
 import org.apache.jena.update.UpdateRequest;
 import org.slf4j.Logger;
+import org.trellisldp.api.Binary;
 import org.trellisldp.api.EventService;
 import org.trellisldp.api.IdentifierService;
 import org.trellisldp.api.MementoService;
@@ -157,9 +158,10 @@ public class TriplestoreResourceService extends DefaultAuditService implements R
 
     @Override
     public Future<Boolean> create(final IRI id, final Session session, final IRI ixnModel, final IRI container,
-                    final Dataset dataset) {
+                    final Binary binary, final Dataset dataset) {
         LOGGER.debug("Creating: {}", id);
-        return supplyAsync(() -> createOrReplace(id, session, ixnModel, container, dataset, OperationType.CREATE));
+        return supplyAsync(() ->
+                createOrReplace(id, session, ixnModel, container, binary, dataset, OperationType.CREATE));
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


core/api/src/main/java/org/trellisldp/api/MutableDataService.java
```diff
@@ -34,10 +34,12 @@
      * @param session the session context for this operation
      * @param ixnModel the LDP interaction model for this resource
      * @param container an LDP container for this resource, {@code null} for none
+     * @param binary metadata for a binary resource, relevant only for ldp:NonRDFSource items: {@code null} for none
```
ajs6f: I know what you mean, but "metadata for a binary resource" seems easily confused with "descriptive triples" (e.g. technical metadata, descriptive metadata). Maybe something as short as `a binary resource representation for {@code ldp:NonRDFSource} items, {@code null} for none`? I know that there is a real distinction between the bitstream, the triples, the info in a `Binary`-- just trying to be clear without being absurdly verbose.

ajs6f: Also, this may not be the PR to do it in, but I think we should start shuffling the `null`able params to the end of the sig? That's my usual style, because they're more "optional"-ish...

components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -205,10 +208,22 @@ private Boolean createOrReplace(final IRI identifier, final Session session, fin
...
             });
         }
+
         // Set the parent relationship
-        if (container != null) {
+        if (nonNull(container)) {
             dataset.add(PreferServerManaged, identifier, DC.isPartOf, container);
         }
+
+        if (nonNull(binary)) {
```
ajs6f: So, post this PR, do _all_ impls have to create these triples from the `Binary`, or is this just what `TriplestoreResourceService` does here because for `TRS`, _everything_ has to be triples?

acoburn: One goal of this PR is to explicitly provide flexibility to an implementation to do whatever it wants with that `Binary` data. The `TRS` puts the data into triples, but that's only because that's what a triplestore would do. I really wouldn't expect any other impl to do that -- and even if it did use triples, it could use whatever triples it wanted to.

core/api/src/main/java/org/trellisldp/api/MutableDataService.java
acoburn: I'll rework the javadoc.

As for the param order, I agree. In fact, I was thinking that moving toward a `Builder` pattern which can more easily handle these "optional-ish" params. I'd like to save that for a separate PR, since this one is really focused on how the `Binary` object gets passed around these interfaces.

ajs6f: A builder could be really nice, but I would want to watch out for the creation of a bazillion short-lived objects. I know that's not as big a problem as once it was, but still. Definitely not for this PR.

components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
ajs6f: 💯

core/api/src/main/java/org/trellisldp/api/MutableDataService.java
acoburn: Sounds good -- I'll move the param order around as part of a separate effort. Then, if a builder still makes sense, it could be added even further down the road.

ajs6f: 👍

PR #129, opened 2018-05-03
Follow File URI specification

### LINKED ISSUE #128: Bad IRI: <file:foo/b> warning in SPARQL
The fuseki log reports this:
```bash
fuseki     | [2018-05-03 07:33:58] SPARQL     WARN  [line: 32, col: 86] Bad IRI: <file:55/1a/ba/551aba6a-5e85-463b-b4a2-c4496bff018b> Code: 57/REQUIRED_COMPONENT_MISSING in AUTHORITY: A component that is required by the scheme is missing.
```

This seems to be caused by the [ViolationRules](https://github.com/apache/jena/blob/master/jena-iri/src/main/java/org/apache/jena/iri/ViolationCodes.java#L621-L645) in Jena IRI that expect a `file:` authority identifier to include 3 slashes. (e.g.  `<file:///foo/b>`).  Note however that It is just a warning, and the query still works.

acoburn:
Resolves #128

Internal file URIs for binary locations have been using
a `file:foo/bar` format, but this is invalid according to RFC 8089,
the "file" URI scheme: https://tools.ietf.org/html/rfc8089

With this change, those internal IRIs are represented as
`file:///foo/bar`.

### DIFF
--- a/components/file/build.gradle
+++ b/components/file/build.gradle
@@ -15,6 +15,7 @@ dependencies {
     implementation("commons-codec:commons-codec:$commonsCodecVersion")
     implementation("commons-io:commons-io:$commonsIoVersion")
     implementation("org.apache.commons:commons-collections4:$commonsCollectionsVersion")
+    implementation("org.apache.commons:commons-lang3:$commonsLangVersion")
     implementation("org.apache.commons:commons-rdf-jena:$commonsRdfVersion")
     implementation("org.apache.jena:jena-osgi:$jenaVersion")
     implementation("org.apache.tamaya:tamaya-api:$tamayaVersion")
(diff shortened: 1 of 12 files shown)


ajs6f:
My feeling about distributed filesystems for Trellis has long been that we (Trellis) should leave distribution to the lower layers. It's hard, it's easy to get things wrong or at least unperformant, and there are lots of good products already out there.

christopher-johnson:
there is the theoretical scenario where one would have to reconstruct a filesystem using the dataset, perhaps to migrate from one data partition to another.  Having a host identifier (that could be an abstract reference) in the file IRI might facilitate that.  That being said, the important part of the IRI for Trellis is not the authority or protocol, but its UUID.  I agree that any distributed identifier abstraction is better handled at a lower level.

ajs6f:
I agree it could be useful to record this info, but I would push to keep it in the user-managed triples, because it will only be important for some sites. I can also think of reasons for which one would specifically want to avoid "leaking" server names, e.g. because of local security policies.

acoburn:
I'd really prefer to keep the `trellis-file` implementation very simple (like it is now). If someone wants to put the `binary` base directory on top of a Ceph cluster, then it's doable but still entirely transparent from the trellis application. For a binary resolver that manages partition (or rack) locations, there ends up being a lot of complexity, and I'm really not sure that the code in this module would even be relevant for what would need to happen there. A fully distributed BinaryService implementation is something I'd like to write at some point, but when I do so, I'd put it all into a completely separate module.

ajs6f:
Big 👍 to keeping that complexity contained. Indeed, for SI, we would be keeping it all the way down beneath Trellis entirely. I have no illusions that I am smarter than the guys who work on GPFS. 😄

PR #138, opened 2018-05-16
adds optional configuration for html template

### LINKED ISSUE #137: Make HTML templates configurable
At present, the code that defines which HTML template is used isn't tied to the application configuration. See [here](https://github.com/trellis-ldp/trellis/blob/master/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisApplication.java#L123) for an example: the `null` should instead be a value that comes from the configuration.

Connecting the template value to the configuration will make it possible for users to more easily adjust the HTML view of a Trellis server. It will also make it easier to include multiple (classpath-based) templates from which to choose.

christopher-johnson:
closes #137

### DIFF
--- a/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisApplication.java
+++ b/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisApplication.java
@@ -11,6 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.trellisldp.app.triplestore;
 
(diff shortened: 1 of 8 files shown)


acoburn:
The code here looks great. The one thing that would be important to add here is some documentation. It appears that there is currently no documentation about any of the asset-related configuration. There is some duplication in terms of where the configuration is documented -- it appears in both `components/app` and `platform/linux`. I'm OK with the duplication, but it means that it would be good to add the docs in both places. In particular, it would be helpful to document how one might make use of the table-based template (and associated javascript dependencies).

PR #142, opened 2018-05-20
Add extended ResourceConfiguration to support BasicAuth for RemoteRDF…

### LINKED ISSUE #141: provide security configuration for remote RDFConnection
I am now looking at how `RDFConnection` works, as research for the future `HttpClient` question.  

I think I have discovered that Fuseki is not able to be secured from RDFConnection, since Shiro is applied in the WebContext environment.  In my test, I locked down Fuseki with a `shiro.ini ` configuration that prevents accessibility via a web request.  However, it seems that the shiro configuration is not applied at the `QueryEngineHttp` API layer, so Trellis (and presumably any Java client) can still access the data freely.  @ajs6f is this correct?  

If Fuseki would accept and process `Authorization` headers from a Java client in `RDFConnection`, then an `HttpClient` could be built in Trellis for the `RDFConnection` from a new expanded `ResourceConfiguration` that might have `BasicAuth` fields for `username` and `password` (in lieu of Fuseki being able to support JWT).

Something like this might work:
```java
if (loc.startsWith("http://") || loc.startsWith("https://")) {
      // Remote
      return create().httpClient(buildHttpClient(config)).destination(loc).build();
}

private static HttpClient buildHttpClient(final TrellisConfiguration config) {
   final Optional<String> userName = ofNullable(config.getResources().getUserName());
   final Optional<String> password = ofNullable(config.getResources().getPassword());
   final CredentialsProvider provider = new BasicCredentialsProvider();
   if (userName.isPresent() && password.isPresent()) {
       final UsernamePasswordCredentials credentials = new UsernamePasswordCredentials(
                    userName.get(), password.get());
      provider.setCredentials(AuthScope.ANY, credentials);
    }
   return HttpClientBuilder.create().setDefaultCredentialsProvider(provider).build();
 }
```

christopher-johnson:
…Connection

closes #141

### DIFF
--- a/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/AppUtils.java
+++ b/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/AppUtils.java
@@ -14,13 +14,7 @@
 
 package org.trellisldp.app.triplestore;
 
-import static java.util.Optional.ofNullable;
 import static javax.jms.Session.AUTO_ACKNOWLEDGE;
-import static org.apache.jena.query.DatasetFactory.createTxnMem;
-import static org.apache.jena.query.DatasetFactory.wrap;
-import static org.apache.jena.rdfconnection.RDFConnectionFactory.connect;
-import static org.apache.jena.rdfconnection.RDFConnectionRemote.create;
-import static org.apache.jena.tdb2.DatabaseMgr.connectDatasetGraph;
 import static org.slf4j.LoggerFactory.getLogger;
 import static org.trellisldp.app.config.NotificationsConfiguration.Type.JMS;
 import static org.trellisldp.app.config.NotificationsConfiguration.Type.KAFKA;
@@ -28,27 +22,18 @@ import static org.trellisldp.app.config.NotificationsConfiguration.Type.KAFKA;
 import io.dropwizard.lifecycle.AutoCloseableManager;
 import io.dropwizard.setup.Environment;
 
-import java.util.Optional;
 import java.util.Properties;
 
 import javax.jms.Connection;
 import javax.jms.JMSException;
 
 import org.apache.activemq.ActiveMQConnectionFactory;
-import org.apache.http.auth.AuthScope;
-import org.apache.http.auth.UsernamePasswordCredentials;
-import org.apache.http.client.CredentialsProvider;
-import org.apache.http.client.HttpClient;
-import org.apache.http.impl.client.BasicCredentialsProvider;
-import org.apache.http.impl.client.HttpClientBuilder;
-import org.apache.jena.rdfconnection.RDFConnection;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.slf4j.Logger;
 import org.trellisldp.api.EventService;
 import org.trellisldp.api.NoopEventService;
 import org.trellisldp.api.RuntimeTrellisException;
 import org.trellisldp.app.config.NotificationsConfiguration;
-import org.trellisldp.app.config.TrellisConfiguration;
 import org.trellisldp.jms.JmsPublisher;
 import org.trellisldp.kafka.KafkaPublisher;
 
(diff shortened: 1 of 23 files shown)


components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/AppUtils.java
```diff
@@ -125,6 +133,18 @@ public static RDFConnection getRDFConnection(final TrellisConfiguration config)
         return connect(createTxnMem());
     }
 
+    private static HttpClient buildHttpClient(final TrellisConfiguration config) {
+        final Optional<String> userName = ofNullable(config.getResources().getUserName());
```
ajs6f: Building an `Optional` here is unidiomatic. Let `getUserName()` return an `Optional`.

components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/AppUtils.java
```diff
@@ -125,6 +133,18 @@ public static RDFConnection getRDFConnection(final TrellisConfiguration config)
...
+        final Optional<String> password = ofNullable(config.getResources().getPassword());
```
ajs6f: Same.

components/app/src/main/java/org/trellisldp/app/config/ResourceConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.trellisldp.app.config;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ResourceConfiguration {
```
ajs6f: This is too generic a name. Something that describes the one single resource configured would be appropriate-- `TriplestoreConfiguration` or the like.

acoburn:
I like the basic idea here, but I'd like to suggest a slightly different approach. The configuration defined in `component/app` is designed to be entirely generic -- the sort of base-level configuration that would apply to all persistence implementations: a cassandra impl, a jdbc impl and a triplestore impl. There will naturally be some implementation-specific configuration options that are needed, and for that, the mechanism currently in place is the `TrellisConfiguration::setAdditionalConfiguration` and `TrellisConfiguration::any`. See [here](https://github.com/trellis-ldp/trellis/blob/master/components/app/src/main/java/org/trellisldp/app/config/TrellisConfiguration.java#L204-L223). And see examples of its use [here](https://github.com/trellis-ldp/trellis/blob/master/components/app/src/test/java/org/trellisldp/app/config/TrellisConfigurationTest.java#L55-L56) and [here](https://github.com/trellis-ldp/trellis/blob/master/components/app/src/test/resources/config1.yml#L94-L95).

So, while I'm certainly no opposed to changing the configuration structure or adding new config options, I'd like to strive for making a distinction between generic configuration and implementation-specific configuration, especially implementation-specific configuration that may not constitute a common use case (here, I'm thinking of the use of a remote triplestore w/ AuthN enabled). If that does become a common use case, then yes, let's reconsider, but until then, I'd like to try to keep the config structure as simple as possible.

Would it be possible to restructure this PR such that the username/password values are set and retrieved via the `setAdditionalConfiguration` mechanism?

One thing to mention is that the test with an embedded triplestore is really excellent -- there are not currently any tests that use a triplestore (they all use either an in-memory dataset or a local TDB2 database).

christopher-johnson:
I agree that this `resources` configuration needs more generic approach.  The reason I started looking at this is to see what could be done to modify/replace the HttpClient in the RDFConnection implementation.  The configuration part in app is actually only for the client, so it is a bit unclear to me how one could extend more configuration options for a client within the `resources` configuration with this first proposal.   I agree with Adam's suggestion to put the ofNullable in the configuration class, and I have done that. 

An idea that has occurred to me is to introduce a generic class in the API called `TriplestoreConnection` .. something like this:
```java
public class RemoteConnection<T> {

    private T connection;

    public void setConnection(T connection) { this.connection = connection; }

    public T getConnection() { return this.connection; }
}
```
This might include a setter and getter for an HttpClient as well.  The configuration in app would then refer primarily to the API defined class rather than the implementation specific RDFConnection.

ajs6f:
That's not a bad idea, but I think @acoburn is pointing out (I missed it but I entirely agree) that all of this should be accessed via `get/setAdditionalConfiguration`, not, as you now have it, from the main config. For those of us who won't be running a triplestore backend, none of this is relevant.

christopher-johnson:
couple of notes on this.
1. using `setAdditionalConfiguration` would mean that the getter needs to be changed to an Optional<Map<String, Object> since the connection related methods need to check if a value `isPresent.`  I can do that or a more direct approach seems to be to contain connection related settings in a new configuration class.  Since all of these settings are optional, a user does not need to care about them, so I do not see any particular advantage in using the additional configuration here.  Up to you, this can still be changed easily.
2. I changed a lot in the TriplestoreResourceServiceTest.  There seemed to be a lot of duplication there, that I tried to reduce.  I cannot understand why the `testRDFConnectionError()` now seems to work so I disabled it.  I would need some help to figure that out.
3. Most of the DatasetConnection changes are semantic.  I do not like the semantics of `RDFConnection` at all.  This is a bad name for an interface, since technically RDF is supposed to be abstract, how can one connect to it?  There are a lot of references to `dataset`, so this seems like a better name for what actually is a `concrete` db connection.  I kept `resources` as the key for the `datasetConnectionConfiguration` because it is basically a convention now, and nothing would be gained by renaming it to `datasetConnection`.  The main idea here is to enable the possibility of different Jena connection implementation types being substituted in TriplestoreResourceService.  This of course would require more changes, but this is just a proposal.

christopher-johnson:
I understand.  There is no urgency in this PR and it can definitely wait.  The username/password configuration is consequential only to one of many possible datasetConnection configurations which is what I would like to expose generically for Trellis.  As such this PR is more of a practical/research exercise than a feature request.

acoburn:
@christopher-johnson what would you like to do with this PR? There is now a better way to write customized app-configuration code, since the `app` component is more flexible in this regard. See, for instance, [this class](https://github.com/trellis-ldp/trellis-ext-db/blob/master/app/src/main/java/org/trellisldp/ext/app/db/AppConfiguration.java) in the `trellis-ext-db` repository, which adds a database connection configuration to the base configuration.

PR #146, opened 2018-06-17
adds Camel Producer notification module

### LINKED ISSUE #145: Add support for camel idempotent repository producer notifications
This is a feature request.  I am interested in the idempotent repository provided with Camel Kafka.  From what I gather, this seems to require a Camel Producer.  A Trellis Camel Producer notification module could then just depend on camel components and not messaging service directly.  The Kafka Idempotent Repository Producer would send messages to Kafka by using a context that used a registry bound to the KafkaIdempotentRepository bean.  In theory, it should be possible to make a generic Camel Producer notification module that could work with any Idempotent Repository variant.

christopher-johnson:
note: the test requires a running Kafka instance, so CI will fail.  Included in resources is a docker-compose for this, but it is not integrated into the test.  This could also be done with an embedded broker and zookeeper.
Other tests can be added for different idempotent repository contexts.

closes #145

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ ext {
     /* Dependencies */
     activationApiVersion = '1.2.0'
     activeMqVersion = '5.15.4'
+    camelVersion = '2.21.1'
     commonsCodecVersion = '1.11'
     commonsCollectionsVersion = '4.1'
     commonsIoVersion = '2.6'
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Sun Jun 17 20:01:03 CEST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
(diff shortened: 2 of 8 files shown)


acoburn:
@christopher-johnson I really like the basic idea here, but I think it might make more sense to put this code into a different project. That is, if the `CamelProducer` were made part of the `trellis-ldp/camel-ldp` project, it could exist as a generic camel-based EventProducer (and it would be _really_ useful there). Then, the kafka-specific parts could be put into the `trellis-ldp/camel-ldp-recipes`. Would that make sense? The general principle here is to keep integration code separate from the main repository, making it easier to release those external components on an independent timeframe.

christopher-johnson:
right, I agree completely.  This is tangential to almost everything in Trellis except for the EventService and an Application configuration.  It is easier to build and maintain as a separate project as well.  I close this here for now.  What do you think about embedding zk and kafka for integration tests?  It is a bit of work and adds dependencies, I guess, but this is how Camel does it.

acoburn:
@christopher-johnson in terms of embedding zk/kafka for integration tests, I'm entirely supportive of that.

PR #152, opened 2018-06-20
Convert from maven to maven-publish plugin

### LINKED ISSUE #149: stop using the gradle maven plugin
The Gradle plugin `maven` is being replaced by `maven-publish`. This means that the `upload` task will be replaced with `publish`. The gradle build script should be updated accordingly.

acoburn:
Resolves #149 

With Gradle 4.8 (we currently use 4.7 for this project), the `maven` plugin becomes deprecated in favor of the `maven-publish` plugin. This PR converts to using the new plugin. The changes primarily relate to signing, POM generation along with a few other items. Because I'm so used to issuing `gradle install` and `gradle upload`, this also aliases the relevant tasks to those commands.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,6 +85,11 @@ ext {
             'trellis-webac',
             'trellis-webapp'
     ]
+
+    omitFromMavenPublishing = [
+        'trellis-linux',
+        'trellis-webapp'
+    ]
 }
 
 allprojects { subproj ->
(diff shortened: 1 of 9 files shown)


acoburn:
The signing behavior is definitely more complex that I was hoping it would be. In time, and with greater familiarity with the plugin, I suspect that there will be opportunities to simplify that (I will experiment a bit with one of the simpler repositories, like `trellis-ext-db`).

acoburn:
It appears that the combination of the `maven-publish` and `signing` plugins will be easier in Gradle 4.8: https://discuss.gradle.org/t/how-to-publish-artifacts-signatures-asc-files-using-maven-publish-plugin/7422/23. That should simplify a lot of this.

PR #164, opened 2018-07-11
Separate MementoService and ResourceService

### LINKED ISSUE #161: Separate the MementoService from the ResourceService
The `ResourceService::getMementos` just proxies calls to a `MementoService`. The two should be separated, as discussed in #148 

Part of this would include removing the `ResourceService::getMementos` method, and updating the HTTP layer to look up memento-related data directly from the `MementoService`.

acoburn:
Resolves #161 

The first commit removes the `ResourceService::getMementos` method.
The second commit removes the `ResourceService::get(IRI, Instant)` method.
The third commit removes the internal use of `MementoService::put` in the `TriplestoreResourceService`

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/MementoService.java
+++ b/core/api/src/main/java/org/trellisldp/api/MementoService.java
@@ -36,6 +36,14 @@ public interface MementoService {
     void put(IRI identifier, Instant time, Stream<? extends Quad> data);
 
     /**
+     * Create a new Memento for a resource.
+     * @param resource the resource
+     */
+    default void put(Resource resource) {
+        put(resource.getIdentifier(), resource.getModified(), resource.stream());
+    }
+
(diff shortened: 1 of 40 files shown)


core/api/src/main/java/org/trellisldp/api/MementoService.java
```diff
@@ -35,6 +35,14 @@
      */
     void put(IRI identifier, Instant time, Stream<? extends Quad> data);
 
+    /**
+     * Create a new Memento for a resource.
+     * @param resource the resource
+     */
+    default void put(Resource resource) {
```
ajs6f: Is this a "snapshotting" operation?

acoburn: Yes, exactly.

PR #174, opened 2018-07-17
Add ServiceBundler interface

### LINKED ISSUE #168: Add an abstraction that collects Trellis services
As suggested in https://github.com/trellis-ldp/trellis/pull/164#pullrequestreview-137204194, it would simplify a lot of method signatures to have a “bundle” sort of object, containing the required services for a trellis application. This is most clearly seen in the HTTP layer, so that may be the best place for this abstraction to reside.

acoburn:
Resolves #168 

This is a deceptively large PR, but there are really only two changes of any consequence, both of which flow from the introduction of a `ServiceBundler` interface: https://github.com/trellis-ldp/trellis/compare/trellis-168?expand=1#diff-0c10dd696fa60291c8319cc82d500a2cR1

1. The `o.t.http.LdpResource` no longer accepts a long list of services but instead accepts a single `o.t.api.ServiceBundler`. This changes a lot of internal code in very trivial ways, but that can't really be avoided, and the changes are, generally, for the best as the method signatures get a lot smaller with this modification.

2. The main application abstraction (e.g. `org.trellisldp.app.AbstractTrellisApplication`) no longer requires implementation of a long list of services; rather it requires implementation of the `ServiceBundler` interface.

One thing to consider here is that, with the exception of `::getAccessControlService`, none of the ServiceBundler methods return `Optional`, basically making all services required (though there are several "no-op" implementations available, if one desires). Does that seem right? Based on how the code uses the service, it seemed like the most natural idiom.

In particular, I am looking for feedback on the `ServiceBundler` interface. This change does not affect any of the other interfaces in the API layer, only shared HTTP-layer code (`trellis-http`) as well as some shared dropwizard-based application code (`trellis-app`).


.appveyor.yml
```diff
@@ -1,8 +1,6 @@
 version: '{build}'
 skip_tags: true
-
-install:
-    - cinst jdk10 -params 'installdir=C:\\jdk10'
+skip_branch_with_pr: true
```
ajs6f: How does this change CI behavior?

components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisServiceBundler.java
```diff
@@ -0,0 +1,150 @@
...
+public class TrellisServiceBundler implements ServiceBundler {
+
+    private static final Logger LOGGER = getLogger(TrellisServiceBundler.class);
+
+    private final MementoService mementoService;
+    private final TriplestoreResourceService resourceService;
+    private final BinaryService binaryService;
+    private final AgentService agentService;
+    private final IOService ioService;
+    private final Optional<AccessControlService> webacService;
```
ajs6f: To me, better `private final  AccessControlService` with a no-op default impl.

components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisServiceBundler.java
```diff
@@ -0,0 +1,150 @@
...
+import org.trellisldp.io.JenaIOService;
+import org.trellisldp.namespaces.NamespacesJsonContext;
+import org.trellisldp.rdfa.HtmlSerializer;
+import org.trellisldp.triplestore.TriplestoreResourceService;
+import org.trellisldp.webac.WebACService;
+
+/**
+ * A triplestore-based service bundler for Trellis.
+ */
+public class TrellisServiceBundler implements ServiceBundler {
```
ajs6f: Should configuration be available from this class?

core/api/src/main/java/org/trellisldp/api/ServiceBundler.java
```diff
@@ -0,0 +1,65 @@
...
+     * Get the {@link MementoService} for the application.
+     * @return the service for interacting with memento resources.
+     */
+    MementoService getMementoService();
+
+    /**
+     * Get the {@link AccessControlService} for the application, if one exists.
+     * @return the service for enforcing access controls.
+     */
+    Optional<AccessControlService> getAccessControlService();
```
ajs6f: Why `Optional` here but not above?

.appveyor.yml
acoburn: This builds and tests all of Trellis on Windows machines. The `skip_branch_with_pr` means that it doesn't build the same thing twice. Removing the jdk10 tests is temporary until the JDK installation process starts working again -- a new JDK 10 version was released yesterday, and the installer packages haven't caught up, causing all the builds to fail, so I'm just removing Windows-based JDK10 tests for now.

core/api/src/main/java/org/trellisldp/api/ServiceBundler.java
acoburn: The idea was that an `AccessControlService` isn't necessary (hence `Optional`), but there are at least two other ways to handle this:

1. Remove the `Optional` and make this just like every other service.
2. Remove this method altogether. This service isn't actually used by the `LdpResource` class, and so, while bundling it with the other services leads to some conveniences w/r/t how the services are instantiated, but it might be better to leave this out of the service bundler.

Personally, I like the second option.

components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisServiceBundler.java
acoburn: Can you clarify this? Do you mean whether the configuration should be accessible via some sort of `::getConfiguration` method?

components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisServiceBundler.java
acoburn: I would agree, though it may be better to simply remove that method.

Issue #148, opened 2018-06-19
Revisit use of Future<?> in API responses


acoburn:
The current ResourceService interface responds with a `Future<>` for certain methods. A `CompletableFuture<>` could be much more flexible. Also, other *Service interfaces could benefit from using `CompletableFuture<>` responses, especially the `BinaryService`.

ajs6f:
We could potentially replace `Optional` with `CompletableFuture` across the API in those methods which connect the backend of choice and the HTTP layer. I'm not sure whether we could do that with `Future` itself very well, tho'. Or at least we would probably go towards runtime exceptions only.

To make this kind of move, I would think we would want to select an _single_ idiom for reporting exceptional execution. With `CompletableFuture`, that would seem to be semantically-precise `Exception`s delivered in the  result itself.

Does that make sense? I really like this [`Maybe`](https://github.com/google/mug#futures) but I don't think we want to use that in the API.

acoburn:
Yes, replacing `Optional` with a `CompletableFuture` seems like it would make a lot of sense. And 👍 to using a single idiom.

How would you suggest modeling a `ResourceService::get` request for a resource that doesn't exist (i.e. `404 Not Found`). The `Optional.empty()` response is a nice fit there, but with `CompletableFuture` are you thinking that a `null` would be returned? Or would it still return an `Optional.empty()` via `CompletableFuture<Optional<?>>`?

ajs6f:
No, it would be a precise `Exception`, like `NoSuchResourceException` or something, that would map directly into a 404. So this
```
class NoSuchResourceException extends RuntimeException { ... }
...
CompletableFuture<Resource> notFound = new CompletableFuture<>();
notFound.completeExceptionally(new NoSuchResourceException());
return notFound;
```
kind of thing, then the caller (say, `trellis-http`) does something like
```
if (result.isCompletedExceptionally()) {
    result.exceptionally(exc -> {
        Throwable ex =  e instanceof ExecutionException || e instanceof CompletionException
                ? exc.getCause()
                : exc;
        if (ex instanceof NoSuchResourceException) {
            // rethrow and let JAX-RS map it into a 404
            throw ex;
        }
        if (ex instanceof SomeOtherReasonException) {
            // rethrow and let JAX-RS map it into something else
            throw ex;
        }
        // something else went wrong
        throw new RuntimeException(ex);
    });
} else {
    // make result into a streaming HTTP response
}
```

ajs6f:
[...]

PR #175, opened 2018-07-19
Initial refactor with CompletableFuture interfaces

acoburn:
This is an incremental PR that includes the simplest parts of a refactor related to #148. In particular, this changes existing `Future` types to `CompletableFuture` types in the `ResourceService` mutating methods. It does not change any of the fetching methods. Changing the `::get` method will result in more significant changes, which will happen separately.

### DIFF
--- a/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
+++ b/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
@@ -54,7 +54,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.Future;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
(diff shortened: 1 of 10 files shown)


ajs6f:
Can you cut a `SNAPSHOT` release for this, to get CI working again downstream? Maybe at some point we should even think about a `SNAPSHOT` repo for Trellis.

acoburn:
I did cut one yesterday evening. You'll need to change these return types to `CompletableFutures` (rather than `Future`): https://github.com/ajs6f/trellis-cassandra/blob/master/src/main/java/edu/si/trellis/cassandra/CassandraResourceService.java#L130-L146

ajs6f:
Ah, right-- the ground shifted under my feet. Thnx

acoburn:
When there are interface changes on `master`, I'm _trying_ to make sure the changes get pushed to the `SNAPSHOT` repo on sonatype. It would be nice to automate this, but I haven't come up with a good way to do this while keeping the sonatype credentials private (and without provisioning an entire server to do that). However, I might be able to pull together an AWS lambda function that provisions a VM, builds the software, runs `./gradlew upload` and then de-provisions the machine.

PR #176, opened 2018-07-19
Add singleton special resources

acoburn:
This is also related to #148 in that it adds two special resource types:
`MISSING_RESOURCE` and `DELETED_RESOURCE`, which will be used
in the HTTP layer.

This also adds a default implementation to the `Resource::hasAcl` method and moves the non-default Resource methods to the top of the class definition.

These resource types will allow us to change the `ResourceService::get` method so that it no longer returns an `Optional<Resource>` but rather a `CompletableFuture<Resource>`.


core/api/src/main/java/org/trellisldp/api/SpecialResources.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        @Override
+        public IRI getInteractionModel() {
+            return null;
+        }
+
+        @Override
+        public Instant getModified() {
+            return Instant.EPOCH;
```
acoburn: Note: it may be better for this method to return `null`.

ajs6f: Definitely. Returning `EPOCH` is letting the computer leak into the real world.

core/api/src/main/java/org/trellisldp/api/SpecialResources.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        @Override
+        public Stream<? extends Quad> stream() {
+            return Stream.empty();
```
ajs6f: Should the `default` in `Resource` be `Stream.empty()`?

core/api/src/main/java/org/trellisldp/api/SpecialResources.java
```diff
@@ -0,0 +1,81 @@
...
+import java.time.Instant;
+import java.util.stream.Stream;
+
+import org.apache.commons.rdf.api.IRI;
+import org.apache.commons.rdf.api.Quad;
+
+/**
+ * Singleton "special" resources.
+ */
+public enum SpecialResources implements Resource {
```
ajs6f: Should this be a static inner type of `Resource`?

core/api/src/main/java/org/trellisldp/api/SpecialResources.java
acoburn: Not a bad idea. In all likelihood, an implementation is still going to need to implement the two `Resource::stream` methods.

core/api/src/main/java/org/trellisldp/api/SpecialResources.java
acoburn: I could go either way on this, but if you'd prefer it to be located in the `Resource` interface, I'll move it.

ajs6f: Yeah, so that it's that much harder to miss them, I would appreciate that.

core/api/src/main/java/org/trellisldp/api/SpecialResources.java
ajs6f: Actually, given that it's actually hard for me to think of an impl that _wouldn't_ need to override this, better to `default Stream<? extends Quad> stream() { throw new UnimplementedOperationException(); }`.

ajs6f: Wait, what am I saying? I need some morning tea. Never mind. Let's leave it unimpled in the interface. I don't know what I'm talking about.

acoburn: Ok, I'll revert that change.

Issue #11, opened 2017-12-29
Add support for asynchronous HTTP


acoburn:
Now that `o.t.api.ResourceService::create` returns a `Future`, it should be possible to enable asynchronous HTTP write operations. At the very least, using [`AysyncResponse`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/container/AsyncResponse.html) could help speed things up.

PR #177, opened 2018-07-27
Convert ResourceService::get to CompletableFuture

acoburn:
Related to #148

This is admittedly a massive PR. I'll let it sit for a few days in case there are comments, but there are, basically, two changes:

The first commit changes the `ResourceService::get` method such that it returns a `CompletableFuture<Resource>` rather than an `Optional<Resource>`. The rest of the changes as part of the commit simply update the code so that it will build and pass all the associated tests.

The second commit is a significant rewrite of the internal plumbing of the HTTP layer to make effective use of `CompletableFuture` objects. These changes prepare the way for implementing #11 (Async HTTP)

After these commits are merged into `master`, I will move on to making the `BinaryService` methods asynchronous (both read and write). Those changes will be minor in comparison.

And after that, the `MementoService` methods will also need to be converted to return `CompletableFuture` objects, but that, too, will be fairly minor.

It is also worth mentioning that the `Resource::isDeleted` method has been removed in these commits, as it is now irrelevant.


core/api/src/main/java/org/trellisldp/api/ResourceService.java
```diff
@@ -136,8 +137,7 @@ default RDFTerm unskolemize(final RDFTerm term) {
      */
     default Stream<? extends Quad> export(final Collection<IRI> graphNames) {
         return scan().map(Triple::getSubject).filter(x -> x instanceof IRI).map(x -> (IRI) x)
-            // TODO - JDK9 optional to stream
-            .flatMap(id -> get(id).map(Stream::of).orElseGet(Stream::empty))
+            .parallel().map(this::get).map(CompletableFuture::join)
```
ajs6f: Do we need `::join` here? Can we not let the parallel threads produce results as they can?

core/http/src/main/java/org/trellisldp/http/impl/PostHandler.java
```diff
@@ -77,108 +86,153 @@
...
-        } catch (final InterruptedException | ExecutionException ex) {
-            LOGGER.error("Error persisting data", ex);
+            binary = null;
         }
 
-        LOGGER.error("Unable to persist data to location at {}", internalId.getIRIString());
-        return serverError().type(TEXT_PLAIN_TYPE)
-            .entity("Unable to persist data. Please consult the logs for more information");
+        // Should this come from the parent resource data?
+        final IRI container = rdf.createIRI(TRELLIS_DATA_PREFIX + getRequest().getPath());
```
ajs6f: Is this assuming something about the relationship of the URIs for container and contained?

core/http/src/main/java/org/trellisldp/http/impl/RdfUtils.java
```diff
@@ -251,7 +247,7 @@ private static Boolean notCompareWithString(final RDFTerm term, final String str
      */
     public static IRI getProfile(final List<MediaType> acceptableTypes, final RDFSyntax syntax) {
         for (final MediaType type : acceptableTypes) {
-            if (RDFSyntax.byMediaType(type.toString()).filter(syntax::equals).isPresent() &&
+            if (RDFSyntax.byMediaType(type.toString()).filter(s -> s.equals(syntax)).isPresent() &&
```
ajs6f: Why this change? Is `syntax` nullable?

acoburn: Yes, there are certain instances in the code where `syntax` will be `null`.

core/api/src/main/java/org/trellisldp/api/ResourceService.java
acoburn: If we continue to return a `Stream<? extends Quad>`, we'll need a `join` somewhere in the chain operations. Performance-wise, it is quite likely that deferring the `join` until the end of the sequence would work. That said, the `export` method is also a bit of a speculative method, given that it is not ever called in the HTTP layer. The idea is to make it available as part of a CLI utility (or some future HTTP endpoint). So until we have an actual use for this (and some data against which to test), I'd be inclined to either leave this as-is or else remove the method entirely; we can always add it back later if there is a use for it.

core/http/src/main/java/org/trellisldp/http/impl/PostHandler.java
acoburn: Good call. This data can be set when the `PostHandler` is initially constructed.

core/api/src/main/java/org/trellisldp/api/ResourceService.java
ajs6f: Dumping all resources at one resource seems profoundly nonRESTful. That kind of work should be taking place on the back end. Let's remove this unless we have a reason for it?

ajs6f: :+1:

PR #178, opened 2018-07-29
Rework BinaryService to use CompletableFuture

acoburn:
Completes #148 

This updates the BinaryService interface to make use of asynchronous processing (via `CompletableFuture`). Changes in this interface result in changes to the `file` and `http` components.

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/BinaryService.java
+++ b/core/api/src/main/java/org/trellisldp/api/BinaryService.java
@@ -17,8 +17,8 @@ import static java.util.Collections.emptyMap;
 
 import java.io.InputStream;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 
 import org.apache.commons.rdf.api.IRI;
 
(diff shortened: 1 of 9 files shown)


core/api/src/main/java/org/trellisldp/api/BinaryService.java
```diff
@@ -36,34 +36,35 @@
...
 
     /**
      * Test whether a binary object exists at the given URI.
      *
      * @param identifier the binary object identifier
-     * @return whether the binary object exists
+     * @return the new completion stage, indicating whether the binary object exists
      */
-    Boolean exists(IRI identifier);
+    CompletableFuture<Boolean> exists(IRI identifier);
```
acoburn: This method can be removed entirely. It is not used anywhere in the codebase, and keeping it here just puts extra burden on implementations of this interface.

core/api/src/main/java/org/trellisldp/api/BinaryService.java
```diff
@@ -72,15 +73,17 @@ default void setContent(IRI identifier, InputStream stream) {
...
+    CompletableFuture<Void> setContent(IRI identifier, InputStream stream, Map<String, String> metadata);
 
     /**
      * Purge the content from its corresponding datastore.
      *
      * @param identifier the binary object identifier
+     * @return the new completion stage
      */
-    void purgeContent(IRI identifier);
+    CompletableFuture<Void> purgeContent(IRI identifier);
```
acoburn: This method is not used anywhere in the codebase. The idea was that, at some point, there would be a mechanism to complete purge this data from the server, either through a new HTTP endpoint or via some CLI utility. For now, I think it would be best to remove the method, and once we're ready to add the aforementioned utility, the method can be re-added.

acoburn:
This is now ready for review. There are three notable items here:

1. The `BinaryService::exists` method has been removed. It is not used in the HTTP layer and there is no need for one to implement this in downstream modules.

2. The `BinaryService::digest` method has been removed. There is a better, existing method: `BinaryService::calculateDigest` which should be used instead. Previously the `::calculateDigest` had a default method referencing `::digest`; that default method has been removed, so downstream modules should adjust accordingly.

3. The use of these methods in the HTTP layer could be optimized. For example, these `BinaryService`-related methods make significant use of the `CompletableFuture::join` method. Some of that is unavoidable, but some could be improved in a subsequent effort. OTOH, making use of this shortcut makes it possible to update the core interface now.

ajs6f:
We're going to see a lot of `::join` until we go fully asynchronous through to the client.

acoburn:
The test code is completely full of `::join` operations, which will likely continue to be the case.

The two places in the HTTP layer where this PR uses `::join` but which could be factored out are [computing digests on GET](https://github.com/trellis-ldp/trellis/pull/178/files#diff-c4f64ac78a8355675c257844ca8c354fR392) and [storing binary data on PUT/POST](https://github.com/trellis-ldp/trellis/pull/178/files#diff-dbdc9e64a5b4bf885ce109b14a499b53R223).

The other place where `::join` is used is in fetching the binary content for use with `StreamingOutput` is [the `GetHandler::getBinaryStream` method](https://github.com/trellis-ldp/trellis/pull/178/files#diff-c4f64ac78a8355675c257844ca8c354fR415). In this case, I tried various ways to make that fully asynchronous, but I ran into issues related to closing the `InputStream`.

PR #187, opened 2018-08-19
Rework exception handling in the HTTP layer

acoburn:
This reworks the exception handling inside the HTTP layer. Rather than creating `Responses` for every exceptional case and passing those along the `CompletableFuture` chain, this make use of the more idiomatic `WebApplicationException` class of exceptions.


acoburn:
@ajs6f this is now ready for review. Here is some additional context for this change:

1. with the async structure of the http methods, there is, at present, no facility for handling cases where an exception (any exception) is thrown in the async flow -- it would lead to a timeout rather than a 500 server error, which is a terrible way to handle errors. That is now fixed with this change.

2. In the current 0.7.x branch, HTTP errors (e.g. 3xx, 4xx and 5xx) are all handled by throwing the standard JAX-RS subclasses of `WebApplicationException`. This actually makes the code considerably simpler, since when an exception happens (even if it's a `412 Precondition Failed`), the irrelevant code doesn't need to (and shouldn't) be executed. It made sense (sort of) to remove that structure as part of the `CompletableFuture` and `AsyncResponse` refactor, but now I am adding these runtime exceptions back in. Since otherwise, the various `*Handler` classes have to manage extra state about whether a request can be short-circuited.

ajs6f:
I'm not sure I understand point 1; what about `CompletableFuture.completeExceptionally(Throwable)` and so forth?

acoburn:
Because we're passing the whole async chain to an `AsyncResponse`, I don't know that it would make sense use `::completeExceptionally`.

The issue w/ point 1 is that currently, there is a async chain such as:

```
resourceService.get(id).thenApply(...).thenApply(...).thenApply(asyncResponse::resume);
```

but if an exception occurs in any of the three `CompletableFuture` stages before calling `AsyncResponse::resume`, then the `::resume` method is never called and the request just hangs.

Adding `::exceptionally(Function<Throwable, T>)` ensures that those exceptions are handled.

ajs6f:
I don't know enough about `AsyncResponse`, so I will read more about it. I think you are saying that we cannot use any method that completes the future because for asynch HTTP, we specifically want to pass up an _un_ completed future? In that event, we're agreed-- as long as exception info is getting passed up via the `CompletableFuture`, I am not objecting. I will read this PR more carefully after reviewing `AsyncResponse`.

acoburn:
Exactly. We are not completing the `Future` inside of the main request thread -- it simply is passed (as a kind of promise) along to the `AsyncResponse` in a separate thread. This way, almost nothing happens in the request thread pool, certainly no disk or network I/O.

ajs6f:
Shall I merge, or squash-and-merge?

acoburn:
Squash-and-merge please. Thanks!

PR #190, opened 2018-08-22
Make async return types consistent

### LINKED ISSUE #183: Inconsistency in async return types
For the asynchronously mutating methods in the API interfaces, there is an inconsistency in the return types. Some methods, such as those in the `ResourceService` interface (`::create`, `::replace`, `::add`, etc), return a `CompletableFuture<Boolean>`, while the `MementoService` and `BinaryService` interfaces (for mutating methods) return a `CompletableFuture<Void>`. At the very least, I would like to make these interfaces more consistent.

In an async context, I don't believe that it makes much sense in the `ResourceService` methods to return a `false` value rather than throwing a runtime exception: there are very nice, generic ways to handle such runtime exceptions, and following that pattern makes sense to me -- as opposed to returning a `false` value.

I would like to propose changing all such (mutating) async method calls so that they return `CompletableFuture<Void>`. Any thoughts on this?

acoburn:
Resolves #183

The changes mutating async methods such that they return
`CompletableFuture<Void>` rather than `CompletableFuture<Boolean>`.

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/ImmutableDataService.java
+++ b/core/api/src/main/java/org/trellisldp/api/ImmutableDataService.java
@@ -34,7 +34,10 @@ public interface ImmutableDataService<T> extends RetrievalService<T> {
      * @param identifier the identifier under which to persist a dataset
      * @param session the session context for this operation
      * @param dataset a dataset to persist
-     * @return the new completion stage
+     * @return a new completion stage that, when the stage completes normally, indicates that the supplied data
+     * were successfully stored in the corresponding persistence layer. In the case of an unsuccessful write operation,
+     * the {@link CompletableFuture} will complete exceptionally and can be handled with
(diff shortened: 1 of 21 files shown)


core/api/src/main/java/org/trellisldp/api/ImmutableDataService.java
```diff
@@ -34,8 +34,8 @@
      * @param identifier the identifier under which to persist a dataset
      * @param session the session context for this operation
      * @param dataset a dataset to persist
-     * @return whether the resource was successfully persisted
+     * @return the new completion stage
```
ajs6f: I think we need a stronger comment than this. We need to explain what returning various values means. E.g. a completed return indicates that persistence  completed, with errors appearing as exceptions in the completed future.

acoburn: Good point. I'll improve the javadocs

PR #191, opened 2018-08-22
A few more tweaks for CDI and some new logging

ajs6f:
The `junit-platform-launcher` stuff is a small fix so that JUnit 5 tests will run in Eclipse.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -126,11 +126,10 @@ allprojects { subproj ->
     }
 
     dependencies {
-        testCompile group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: "${junitVersion}"
-        testCompile group: 'org.junit.platform', name: 'junit-platform-launcher', version: "${junitLauncherVersion}"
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -125,6 +126,8 @@ allprojects { subproj ->
     }
 
     dependencies {
+        testCompile group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: "${junitVersion}"
```
acoburn: I think the `junit-jupiter-engine` is already included at the "runtime" scope. Does it need to be at the compile scope instead?

ajs6f: Well, I don't know for sure-- I don't pretend to understand what's going on here very well. I could not run the tests, I searched a bit for similar cases and got this from Stack Overflow, and it worked for me. If you want to leave it out I can do something local to workaround, certainly.

acoburn: In my local environment, at least, I can remove that line and the tests all run. I'd leave the `junit-platform-launcher` part in place, but if removing that line doesn't work for you, then leave it in.

PR #198, opened 2018-08-31
Add backend initialization code

### LINKED ISSUE #196: Trellis should initialize its own backend
To the extent that any particular predefined information must be present in the repo backend for proper operation, `trellis-http` should take responsibility for inserting it if not present at startup time.

acoburn:
Resolves #196 

This is a first stab at having the HTTP layer auto-initialize the backend. The triplestore impl still initializes itself independently, but that can be changed if this seems like the right approach.

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/LdpResource.java
+++ b/core/http/src/main/java/org/trellisldp/http/LdpResource.java
@@ -140,7 +140,14 @@ public class LdpResource implements ContainerRequestFilter {
     }
 
     /**
-     * Initialize the Trellis backend.
+     * Initialize the Trellis backend with a root container and default ACL quads.
+     *
+     * @apiNote In a CDI context, this initialization step will be called automatically.
+     *          In a Java SE context, however, it may be necessary to invoke this method
+     *          in code, though a ResourceService implementation may still choose to
+     *          initialize itself independently of this method. In either case, if the
+     *          persistence backend has already been initialized with a root resoure and
+     *          root ACL, this method will make no changes to the storage layer.
      */
     @PostConstruct
     public void initialize() {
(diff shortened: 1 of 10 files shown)


core/http/src/main/java/org/trellisldp/http/LdpResource.java
```diff
@@ -128,6 +139,37 @@ public LdpResource(final ServiceBundler trellis, final String baseUrl) {
         this.trellis = trellis;
     }
 
+    /**
+     * Initialize the Trellis backend.
+     */
+    @PostConstruct
+    public void initialize() {
```
ajs6f: In a non-CDI context, how does this get called? Does the responsibility lie with the backend?

acoburn: I assume that it would be up to the wiring system. So, in the Dropwizard wiring, I'd just call `initialize()` right in the code.

ajs6f: Okay, can we get an `@apiNote` to that effect, or other doc? I'm a little worried about folks not realizing that they may or may not need to take action depending on their choice of wiring toolbox.

acoburn: Ah yes, definitely

acoburn:
@ajs6f if this looks good to you, I think this can be merged.

ajs6f:
Can I put review off until tomorrow? Heck of a day today...

acoburn:
Of course! there's no rush at all.

ajs6f:
Thanks!

acoburn:
@ajs6f would you still like to review this PR, or shall I just merge it?

PR #204, opened 2018-09-18
Move event service logic into HTTP layer

### LINKED ISSUE #201: Move EventService into HTTP layer
Notification events are, at present, emitted from within the backend persistence layer, which means that each persistence implementation needs to implement its own logic for emitting notifications. By moving that logic into the HTTP layer, it should be possible to generalize that behavior. This change would also put the code into a good position for supporting WebSocket-based events as in #29.

Before implementing a particular solution here, it seems that the two most obvious paths for implementation would be to either:

1. Change the mutating `ResouceService` methods so that they return a `Stream<Event>` (in this way, the persistence impl keeps track of the resources that were changed -- it gets more complicated w/ direct and indirect resources), and the HTTP layer simply emits events for whatever `Event`s were returned by the `ResourceService`.
2. Don't change the `ResourceService` interface at all, and just use the existing public methods to determine which resources changed.

I have a preference for the second option, provided that such a design is not overly cumbersome to implement (the first method would be pretty trivial to implement, but it would require all persistence layers to implement their own notification logic).

acoburn:
Resolves #201

### DIFF
--- a/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisServiceBundler.java
+++ b/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisServiceBundler.java
@@ -57,6 +57,7 @@ public class TrellisServiceBundler implements ServiceBundler {
     private final BinaryService binaryService;
     private final AgentService agentService;
     private final IOService ioService;
+    private final EventService eventService;
 
     /**
      * Create a new application service bundler.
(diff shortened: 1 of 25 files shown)


acoburn:
Just a note that this change adds a method to the `org.trellisldp.api.ServiceBundler` interface, namely: `EventService getEventService()`. Otherwise, there are no relevant API-level changes.

PR #213, opened 2018-09-26
Refactor ResourceService::delete method

### LINKED ISSUE #212: Clean up ResourceService::delete method
The `ResourceService::delete` method has three parameters: IRI identifier, IRI interactionModel, Dataset dataset.

The `interactionModel` and `dataset` properties could probably be removed (I don't have any use for them), and a `container` property could be added to bring this into alignment with `::create` and `::replace`.

acoburn:
Resolves #212


ajs6f:
Ok, I don't have very strong feelings about this.

PR #215, opened 2018-09-27
Support PUT-uncontained for creation requests

### LINKED ISSUE #214: Trellis assumes that IRI hierarchy represents extant containment hierarchy for new resources
Mouthful of a title there.

Trellis (in the HTTP layer) assumes that if a resource that it is creating from `PUT` has an IRI that features a more leftward segment, it exists as a container. E.g. `http://example.com/trellis/somestuff/myresource` shows
`http://example.com/trellis/somestuff` as a "subterm" and Trellis then assumes that `http://example.com/trellis/somestuff` exists (from the IRI) and assigns it as a container to `myresource`, using the normal call it makes to `ResourceService::create`. But `somestuff` is not there and a `GET` issued there returns 404 because it was never actually created.

We can play this several ways. Some people would like to do "`mkdir -p`". I think that's terribly wrong. I would rather be able to use `PUT` to make uncontained resources. In any event, the current behavior can't be right. Either we should actually create `somestuff` or not.

acoburn:
Resolves #214

I'd like to add some more tests, but this is a start.

### DIFF
--- a/components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java
+++ b/components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java
@@ -289,7 +289,8 @@ public interface LdpBasicContainerTests extends CommonTests {
             final Graph g = readEntityAsGraph(res.getEntity(), getBaseURL(), TURTLE);
             final IRI identifier = rdf.createIRI(getContainerLocation());
             assertAll("Check the resulting graph", checkRdfGraph(g, identifier));
-            assertTrue(g.contains(identifier, LDP.contains, rdf.createIRI(child4)), "Check for an ldp:contains triple");
+            assertFalse(g.contains(identifier, LDP.contains, rdf.createIRI(child4)),
+                    "Check for an ldp:contains triple");
             setThirdETag(res.getEntityTag());
             assertTrue(getThirdETag().isWeak(), "Check for a weak ETag");
             assertNotEquals(getFirstETag(), getThirdETag(), "Check ETags 1 and 3");
--- a/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResource.java
+++ b/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResource.java
@@ -17,6 +17,8 @@ import static java.util.Objects.nonNull;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
 import static java.util.stream.Stream.builder;
 import static java.util.stream.Stream.concat;
+import static java.util.stream.Stream.empty;
+import static java.util.stream.Stream.of;
 import static org.apache.jena.graph.NodeFactory.createURI;
 import static org.apache.jena.graph.Triple.create;
 import static org.slf4j.LoggerFactory.getLogger;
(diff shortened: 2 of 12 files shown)


components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java
```diff
@@ -289,7 +289,8 @@ default void testCreateContainerViaPut() {
             final Graph g = readEntityAsGraph(res.getEntity(), getBaseURL(), TURTLE);
             final IRI identifier = rdf.createIRI(getContainerLocation());
             assertAll("Check the resulting graph", checkRdfGraph(g, identifier));
-            assertTrue(g.contains(identifier, LDP.contains, rdf.createIRI(child4)), "Check for an ldp:contains triple");
+            assertFalse(g.contains(identifier, LDP.contains, rdf.createIRI(child4)),
```
ajs6f: Just as a side note, this test is failing for me now, I think it will pass after this PR, yay!

components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResource.java
```diff
@@ -177,6 +179,16 @@ protected void fetchData() {
         });
     }
 
+    @Override
+    public Optional<IRI> getContainer() {
+        return graph.stream(identifier, DC.isPartOf, null).map(Triple::getObject).flatMap(obj -> {
+            if (obj instanceof IRI) {
```
ajs6f: Hm. This algo is a bit odd to me, but I probably don't understand the 3store impl well enough. Are we saying, the first triple with an `isPartOf` predicate, take its object and go?

core/api/src/main/java/org/trellisldp/api/Resource.java
```diff
@@ -119,6 +129,17 @@ public String toString() {
      */
     Instant getModified();
 
+    /**
+     * Get the container for this resource.
+     *
+     * @apiNote returning an empty Optional should indicate here that the resource is not
+     *          contained by any parent resource. This may be because it is a root resource
+     *          and therefore not contained by any other resource. Alternately, it could also
```
ajs6f: I think "alternatively" is to be preferred.

components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResource.java
acoburn: The triplestore impl has an internal data structure where each resource has a pointer to its parent, and it uses the `dc:isPartOf` predicate for that. Here we're just exposing that through the `getContainer` method. The `instanceof` check is just there for completeness -- the value of the `obj` should always be an `IRI` (Triple::getObject returns `RDFTerm`)

ajs6f: So could I write:
```
graph.stream(identifier, DC.isPartOf, null).map(Triple::getObject).filter(IRI.class::isInstance).map(IRI.class::cast).findFirst();
```
? Just trying to make sure I understand the 3store impl, since it is kind of "reference-impl-ish".

acoburn: Yes, that would be a much nicer formulation.

acoburn:
@ajs6f don't merge quite yet -- I have some more tests I'd like to run. Particularly, I'd like to verify that `PUT` for creation doesn't update the modified date of the parent container and that no update notification is sent out for the parent.

ajs6f:
NP-- just give me a ping for further review/merge-age.

acoburn:
@ajs6f I'm now satisfied with this. Feel free to merge if you like the last commit. I verified that `PUT` for creation does not change the modification date or ETag for the "parent" container. Nor are there events sent out for that container.

ajs6f:
We're good now, right?

acoburn:
Yes, if you're happy with this feel free to merge.

PR #220, opened 2018-10-04
Move authentication into its own set of modules

### LINKED ISSUE #216: Move Authentication processing into its own module
At present (and for reasons of expediency) all of the authentication processing is handled by Dropwizard and is located in the `trellis-app` module. In order to generalize the AuthN processing and make it available to non-Dropwizard-based packaging, this code should be moved into a JAX-RS filter in the HTTP layer. The code just converts JWT bearer tokens into `java.security.Principal` objects, so it shouldn't be too difficult to relocate these components.

acoburn:
Add JWT-based Authentication filter
Add Basic Authentication filter
Remove dropwizard-based auth

Resolves #216


ajs6f:
This is a big 'un. The idea here is that after this changeset, we should be able to use nothing more than standard JAX-RS (plus whatever the particular deployment uses, e.g. Spring or CDI or whatever) to enable authN, but if we don't, we simply won't have any and the repo will work as usual, correct? (Obviously modulo WebAC not doing anything because there are no agents to work with.)

acoburn:
That is exactly correct.

PR #226, opened 2018-10-06
Separate Trellis resource matcher and request filter

### LINKED ISSUE #203: Factor filter method out of LdpResource into its own class
The `ContainerRequestFilter` part of the `LdpResource` class should be moved into its own class. See conversation [here](https://github.com/trellis-ldp/trellis/issues/202#issuecomment-422017770)

acoburn:
Resolves #203 

This just extracts the `ContainerRequestFilter` portions of the current `TrellisHttpResource` class into its own class.


acoburn:
Yes, that is exactly the purpose of this filter. In addition, it will throw exceptions that correspond to what is defined by those standards, whether that's a `405 Not Allowed` or a generic `400 Bad Request`. This way, that logic doesn't have to be duplicated inside each of the `PUT`, `POST`, etc handlers.

acoburn:
The other thing to mention about this is that this filter processes requests before the persistence layer is ever called, so it can short-circuit these error conditions proactively.

PR #229, opened 2018-10-08
Put all delete logic into a single test

### LINKED ISSUE #219: Test code oddity?
I'm implementing the `LdpBasicContainerTests` for `trellis-cassandra` -- only one that relates directly to containment is still broken: `testDeleteContainer`. After a lot of logging, I noticed that the IRIs being used for container and child look odd. Then I looked at the tests that are passing, e.g. `testCreateContainerViaPost` and noticed something different about them. At the [corresponding assertion](https://github.com/trellis-ldp/trellis/blob/master/components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java#L260) the IRI of the child (that we want to see in the object slot of an `ldp:contains` triple) is a reference that [we set above](https://github.com/trellis-ldp/trellis/blob/master/components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java#L249) whereas in the still-failing test we use the `getChildLocation()` method [directly](https://github.com/trellis-ldp/trellis/blob/master/components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java#L348)…

Is that right? Is there a reason that we're not taking a reference at the top of the test? I could be totally nuts here, but it's really strange to me to _that_ assertion failing in _that_ test, because no deletion has yet occurred at that point and the more-or-less the same test code seems to be passing nicely in the other tests.

acoburn:
Resolves #219 

If this change fixes the issue described, then I will add one more commit, cleaning up some of the now-unneeded methods.

### DIFF
--- a/components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java
+++ b/components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java
@@ -340,26 +340,34 @@ public interface LdpBasicContainerTests extends CommonTests {
         final RDF rdf = getInstance();
         final EntityTag etag;
 
+        // TODO remove the getChildLocation and setChildLocation methods from this interface
+        final String childResource;
+
+        try (final Response res = target(getContainerLocation()).request().post(entity("", TEXT_TURTLE))) {
+            assertAll("Check for an LDP-RS", checkRdfResponse(res, LDP.RDFSource, null));
+            childResource = res.getLocation().toString();
+        }
+
+        meanwhile();
+
         try (final Response res = target(getContainerLocation()).request().get()) {
             assertAll("Check for an LDP-BC", checkRdfResponse(res, LDP.BasicContainer, TEXT_TURTLE_TYPE));
             final IRI identifier = rdf.createIRI(getContainerLocation());
             final Graph g = readEntityAsGraph(res.getEntity(), getBaseURL(), TURTLE);
             assertAll("Verify the resulting graph", checkRdfGraph(g, identifier));
-            assertTrue(g.contains(identifier, LDP.contains, rdf.createIRI(getChildLocation())),
+            assertTrue(g.contains(identifier, LDP.contains, rdf.createIRI(childResource)),
                     "Check for the presence of an ldp:contains triple");
             etag = res.getEntityTag();
             assertTrue(etag.isWeak(), "Verify that the ETag is weak");
         }
 
-        meanwhile();
-
         // Delete one of the child resources
-        try (final Response res = target(getChildLocation()).request().delete()) {
+        try (final Response res = target(childResource).request().delete()) {
             assertEquals(SUCCESSFUL, res.getStatusInfo().getFamily(), "Check the response type");
         }
 
         // Try fetching the deleted resource
-        try (final Response res = target(getChildLocation()).request().get()) {
+        try (final Response res = target(childResource).request().get()) {
             assertEquals(CLIENT_ERROR, res.getStatusInfo().getFamily(), "Check for an expected error");
         }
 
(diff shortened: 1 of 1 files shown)


ajs6f:
Oddly, this doesn't build for me:
```
> Task :trellis-auth-basic:compileJava FAILED
/Users/ajs6f/Documents/trellis/trellis/auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java:99: error: unreported exception X; must be caught or declared to be thrown
            .map(credentials -> authenticate(credentials).orElseThrow(() -> new NotAuthorizedException(challenge)))
                                                                     ^
  where X,T are type-variables:
    X extends Throwable declared in method <X>orElseThrow(Supplier<? extends X>)
    T extends Object declared in class Optional
1 error
```
Are you working on something other than Java 8?

acoburn:
I mostly use jdk11 locally, but it gets built on jdk8 with CI. I'll try locally with jdk8

acoburn:
Strange. This module builds for me with jdk8
```
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)
```

ajs6f:
Very weird. I'm on a train, so I'll look at this later today.

acoburn:
if there's some change that will get this to build for you, feel free to send over a PR. If you send it to `master`, I'll rebase that into this branch.

components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java
```diff
@@ -340,34 +340,42 @@ default void testDeleteContainer() {
...
-
         // Delete one of the child resources
-        try (final Response res = target(getChildLocation()).request().delete()) {
+        try (final Response res = target(childResource).request().delete()) {
             assertEquals(SUCCESSFUL, res.getStatusInfo().getFamily(), "Check the response type");
         }
 
         // Try fetching the deleted resource
-        try (final Response res = target(getChildLocation()).request().get()) {
+        try (final Response res = target(childResource).request().get()) {
```
ajs6f: I've finally gotten this test to run, but only by putting a `meanwhile();` above this line. You can doubtless see why-- the Cassandra backend is fully async and without the `meanwhile();` this section of test runs "too quickly" after the delete and still returns a 200.

We have a long-term problem with testing async backends, but `meanwhile()`, can we move forward with a `meanwhile();`? 😀

ajs6f: Hm, got to take this back a bit-- the `meanwhile();` only works some of the time, not surprisingly.

ajs6f: There was an underlying problem in my code, and we don't need any `meanwhile();` here.

acoburn:
@ajs6f I added a commit here that isolates all ETag checks. There is still a potential synchronization issue (e.g. with a C* backend), and so it may be necessary to replace all those `meanwhile()` calls with an `awaitility` method to wait for the resource/container to change. But give this a try and let me know if it works for you.

components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java
```diff
@@ -283,17 +272,18 @@ default void testCreateContainerWithSlug() {
...
 
+        final EntityTag initialETag = getETag(getContainerLocation());
+
         try (final Response res = target(getContainerLocation()).request().post(entity("", TEXT_TURTLE))) {
             assertAll("Check for an LDP-RS", checkRdfResponse(res, LDP.RDFSource, null));
             childResource = res.getLocation().toString();
         }
 
-        meanwhile();
+        await().until(() -> !initialETag.equals(getETag(getContainerLocation())));
```
ajs6f: Is there a timeout on this? I worry a bit about tests that hang indefinitely... but maybe it's not an issue?

acoburn: The default is 10 seconds. After that, the test will fail.

ajs6f:
Okay, the `awaitility`-powered tests now timeout, which seems like a good sign that something is going on with my backend, not the tests. Am I correct to assume that in `LdpBasicContainerTests` there is now _no_ cross-test ETag interaction? That's what it looks like, and I think that proves the wrongness of my code.

acoburn:
Correct, there is no cross-test ETag interaction. The _only_ cross-test interaction is fetching the location of the basic container in which this group of tests runs. The value is set in the `BeforeAll` section, and after that everything just reads from that value.

ajs6f:
Okay, good, in the sense that we're starting to get some certainty about this stuff. I'm going to go away for a bit and inspect my `Resource` impl very hard. I have JSON-LD WG today, so you may not hear from me until the afternoon Eastern time.

PR #236, opened 2018-10-11
Making things very explicit for less-intelligent compilers

ajs6f:
The two things going on here are:

1. Adding explicit type params for `orElseThrow`, which really we oughtn't have to, but I'm not going to spend my life arguing with compilers.
2. Making some anon classes static to avoid some GC (I know some RTEs are smart enough to do this themselves, but TRUST NO ONE.)


ajs6f:
Looks like Appveyor beefed on dependencies, but I don't think I can rerun it.

acoburn:
Appveyor can be pretty flaky -- in fact, I just ignore this category of build failures. What's important is that the actual tests don't fail when there's no other networking-based failure going on.

acoburn:
For some reason, github didn't seem to add this commit to master. I'll do so manually.

acoburn:
Of course -- you target the PR with this commit. Nevermind.

Issue #237, opened 2018-10-12
ETag and ordering in tests


ajs6f:
Reading [LdpBasicContainerTests](https://github.com/trellis-ldp/trellis/blob/master/components/test/src/main/java/org/trellisldp/test/LdpBasicContainerTests.java), I see many assertions where various ETags are compared, using `getFirstETag()`, `getSecondETag()`, etc. But in each test, only one method of the `setSecondETag(...)` group is being called. How are the other values being recorded? Is this a case in which the tests have to run in a certain order?

acoburn:
Notwithstanding the naming of those methods, the order is unimportant (at least, it _should_ be unimportant -- I could be missing something). They are each ETags that are gathered/retrieved at different points of the test execution. The important thing being tested is that the particular ETag under evaluation is unique w/r/t the other ETags (whether they have been set or not).

ajs6f:
I'm still a bit confused: when you write "at different points of the test execution" you mean "at different points within a single test"? Or "at different points as various tests are executed"?

acoburn:
I mean "at different points as various tests are executed". So as test A runs, it will record its ETag and verify that it is different than the other ETags (all of which may be `null` if that test is run first). When test B runs (if the expectation is that the ETag changes), that ETag will be compared to the others, confirming that it is different. It shouldn't matter what order A and B are run in.

ajs6f:
What if they run in parallel?

ajs6f:
I should add some actual data here, I suppose 😁 . What I'm seeing is:
```
[ERROR] Failures:
[ERROR]   BasicContainmentIT.testCreateContainerViaPost Check that the ETag has been changed ==> expected: not equal but was: <W/"8c93f5c662d5fd86f28c57b0ce9beed3">
[ERROR]   BasicContainmentIT.testCreateContainerViaPut Check ETags 3 and 4 ==> expected: not equal but was: <W/"8c93f5c662d5fd86f28c57b0ce9beed3">
[ERROR]   BasicContainmentIT.testDeleteContainer Verify that the ETag value is different ==> expected: not equal but was: <W/"8c93f5c662d5fd86f28c57b0ce9beed3">
```
which is why I'm trying to figure out the "whens" of this thing.

acoburn:
That points to a need to make the tests stronger.

acoburn:
[...]

PR #238, opened 2018-10-12
Fix for219

ajs6f:
I'm reoffering this commit, because the current `master` doesn't build for me for the exact same reasons as didn't `trellis-219`, so whatever happens with https://github.com/trellis-ldp/trellis/pull/229, I would like to get this into `master`.


ajs6f:
Urg, just realized I took _your_ commit with mine. I meant to just cherry-pick mine, but actually, there's nothing wrong with yours, I don't think. So merging either this 'un or #229 would let me get back on the road to look at #237 .

acoburn:
How about if I just cherry-pick this commit into master? I agree that either way, it should make its way into the master branch.

ajs6f:
Totally cool with me. That should let me build and roll forward.

acoburn:
Fixed via https://github.com/trellis-ldp/trellis/commit/87038289581aa263ca6c06d1552491008a518324

PR #247, opened 2018-10-16
Add Javadoc that initialization matters

ajs6f:


acoburn:
@ajs6f I just added a commit here that will help with JDK 11 environments. I also added the `@implSpec` javadoc tag.

ajs6f:
Great, let's just make sure the Javadoc build has the right param to use that on Java8.

acoburn:
If you like the way that looks now, feel free to merge.

PR #255, opened 2018-10-22
Add docker configuration

acoburn:
This adds a configuration for creating a docker container for Trellis and pushing that container to dockerhub.

### DIFF
--- a/platform/linux/build.gradle
+++ b/platform/linux/build.gradle
@@ -1,6 +1,7 @@
 apply plugin: 'application'
 apply plugin: 'distribution'
 apply plugin: 'nebula.ospackage-application'
+apply plugin: 'com.palantir.docker'
 
 description = 'Trellis Triplestore Application'
 mainClassName = 'org.trellisldp.app.triplestore.TrellisApplication'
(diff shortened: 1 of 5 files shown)


acoburn:
@gregjan would you be able to look at this Docker configuration? I cribbed from your addition to `trellis-ext-db`.

platform/linux/build.gradle
```diff
@@ -61,6 +62,22 @@ buildDeb {
     preUninstall file('src/deb/preUninstall.sh')
 }
 
+docker {
+    name "trellisldp/trellis:${project.version}"
+    tags 'latest'
+    dockerfile file('src/docker/Dockerfile')
+    files tasks.distTar.outputs, './src/dist/etc/config.yml','./src/docker/command.sh'
```
gregjan: I think this line must be revised. It refers to the outputs of the "distTar" task, which I am pretty sure does not exist in stock trellis, at least I don't see it here.. The error I get upon Docker startup is that it cannot find "trellis" on the path.. "trellis" would have been packaged at /opt/trellis/bin/trellis if things had worked correctly.

acoburn:
@gregjan Thanks! Yes, adding a console logger makes a lot of sense. I'd welcome any suggestions to improve on this, either via PR or just a pointer to working code/configuration.

platform/linux/build.gradle
acoburn: Ok, the `distTar` task _should_ be present here, but I'll need to look at this more carefully.

PR #258, opened 2018-10-23
Adjust injection constructor on IO service

acoburn:
This moves the injection point for the Jena-based IO Service to the no-arg constructor.

@ajs6f is this a good idea? Would this make things easier or harder for CDI? We can certainly leave this as it is, if that would be preferable; otherwise, I'd like to start to favor the no-arg constructors for dependency injection.


ajs6f:
The more no-arg ctors generally (all the way down the system) the better.

PR #260, opened 2018-10-24
Push snapshot artifacts to Sonatype

acoburn:
This will likely need some additional adjustment, but this should be what's needed for pushing SNAPSHOT artifacts to sonatype.


acoburn:
Resolved via 9e19c28979f0d606fc840d505b084fda8a4f7ceb

PR #262, opened 2018-10-25
Add ResourceService::touch method

### LINKED ISSUE #257: Should backends be altering the modified dates of containers?
When a resource is created, the modified date of any container it happens to possess must be altered, because ETags rely on that date to sense change. Currently, this is the responsibility of the backend. 

I'm wondering whether this is a good workflow, because it means that a request to create a resource affects one resource in the HTTP layer, but two in the backend. That seems to go a bit against Trellis' usual simplicity. We haven't yet adumbrated any principle about this and it's not clear to me what it would be in general, but in this case would it be better for the HTTP to make two requests of the backend? One to create the resource, the other to touch the container? 

Another question: should the backend (or HTTP layer) touch the container when a child is altered, or only when created? I'm pretty sure we thought this through in  LDP terms quite a while ago, but it's been too long...

acoburn:
Resolves #257 

This adjusts the core interfaces to introduce a `ResourceService::touch` method, and applies the changes to the HTTP layer.

The first commit adds the method and makes use of it in the HTTP layer.
The second commit adds an implementation in the triplestore persistence layer.

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/ResourceService.java
+++ b/core/api/src/main/java/org/trellisldp/api/ResourceService.java
@@ -121,9 +121,9 @@ public interface ResourceService extends MutableDataService<Resource>, Immutable
     /**
      * Update the modification date of the provided resource.
      *
-     * @apiNote there may be cases where the resource with the given identifier does not (yet) exist.
-     *          In those cases, the implementation should return a completed future as if the resource
-     *          existed. Exceptions should be thrown only if there is an error updating an existing resource.
+     * @apiNote In general, when this method is called, it can be expected that the target resource
+     *          will already exist. In cases where that resource does not already exist, an implementation
+     *          may choose to throw an exception, log an error or simply do nothing.
      * @param identifier the identifier of the resource
      * @return a new completion stage that, when the stage completes normally, indicates that the
      *         identified resource has been updated with a new modification date.
--- a/core/api/src/main/java/org/trellisldp/api/ResourceService.java
+++ b/core/api/src/main/java/org/trellisldp/api/ResourceService.java
@@ -20,6 +20,7 @@ import static org.trellisldp.api.RDFUtils.getInstance;
 
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 
 import org.apache.commons.rdf.api.BlankNode;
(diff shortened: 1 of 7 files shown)


core/api/src/main/java/org/trellisldp/api/ResourceService.java
```diff
@@ -117,6 +118,18 @@ default RDFTerm unskolemize(final RDFTerm term) {
         return term;
     }
 
+    /**
+     * Update the modification date of the provided resource.
+     *
+     * @apiNote there may be cases where the resource with the given identifier does not (yet) exist.
+     *          In those cases, the implementation should return a completed future as if the resource
```
ajs6f: Wait, whoa-- if there is no resource, this method returns as though there was? I guess that's not incorrect, but it seems odd... is there an impl reason for this?

I hate to say it, but if this returned `CompletableFuture<? extends Resource>` we could return `MISSING_RESOURCE` in that case.

acoburn: Actually, I wrote that note before wiring it up in the HTTP layer (and before writing the triplestore implementation), and in practice, there will always be an existing resource. There _may_ be some corner cases, such as a child resource is added to a direct container at the same time that the associated membership resource gets deleted, but that may be outside the scope of what Trellis manages.

Basically, the way this is wired in the HTTP layer follows the same logic that generates notifications, which involves some level of verification that the parent/membership resources exist in the first place.

Which is to say, I should really revise or remove that note. What I was trying to express is that, given that Trellis doesn't enforce a transactional model, it is theoretically possible that the resource wouldn't exist. However, if an implementation wanted to throw an exception in the case of a non-existing resource, that wouldn't necessarily be wrong.

ajs6f: Okay, this:

>  if an implementation wanted to throw an exception in the case of a non-existing resource, that wouldn't necessarily be wrong.

makes me feel better, and if it's okay, that's what I intend to do in `trellis-cassandra`.

acoburn: I've updated the javadoc for this, which should clarify things.

acoburn:
And the new SNAPSHOT artifacts are already on sonatype thanks the the new configuration in Travis-CI. Automation FTW!

Issue #264, opened 2018-10-31
ResourceService::getContainer is synchronous


ajs6f:
We have `Optional<IRI> getContainer(final IRI id)` and as far as I see, that is the sole remaining synchronous method in `ResourceService` (that actually deals with resources). Do we want to go to `CompletableFuture<IRI> getContainer(final IRI id)` or even `CompletableFuture<Resource> getContainer(final IRI id)` and [finish the job](https://www.youtube.com/watch?v=avp836kKCjY)?

acoburn:
I looked through the code and there are really only two uses of this method:

  * Once in the [TrellisHttpResource](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java#L297-L301)
  * A few times in the [WebACService](https://github.com/trellis-ldp/trellis/blob/master/components/webac/src/main/java/org/trellisldp/webac/WebACService.java#L167)

AFAICT, all of these methods get the container IRI and then immediately fetch the corresponding resource (if one exists). Which suggests that we might as well change this to a `CompletableFuture<Resource>`.

ajs6f:
I'll try to make a PR sometime this week.

acoburn:
One potential issue here is this: in order to perform WebAC authorization, we need to recursively look up the parent resource if no ACL is defined on the resource in question. Here, there is a bit of a conflict between the SOLID webacl definition and the ability in Trellis to create uncontained resources. In order to follow the SOLID guidelines, the WebAC module will follow a structural-logical containment hierarchy based on the resource IRI regardless of the existence of a resource: so `/foo/bar/baz` is _ipso facto_ contained by `/foo/bar`, even if `/foo/bar` doesn't exist. That hierarchy is followed until either (a) an _extant_ resource contains an ACL definition or (b) we've reached the root resource. SOLID also _requires_ that the root resource have an ACL. We could treat every uncontained resource as its own distinct root resource, but then do we _require_ that an ACL is defined on each of those resources? That seems a little too prescriptive and/or cumbersome (it also gets into a chicken-egg issue).

[...]

PR #265, opened 2018-10-31
Move getContainer logic inside the WebAC service

acoburn:
Related to #264 

This sets the stage for reworking the ResourceService::getContainer method.

### DIFF
--- a/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
@@ -164,7 +164,7 @@ public class WebACService implements AccessControlService {
         final Set<IRI> modes = getModesFor(identifier, agent);
         // consider membership resources, if relevant
         if (checkMembershipResources && hasWritableMode(modes)) {
-            resourceService.getContainer(identifier).map(resourceService::get).map(CompletableFuture::join)
+            getContainer(identifier).map(resourceService::get).map(CompletableFuture::join)
                 .flatMap(Resource::getMembershipResource).map(WebACService::cleanIdentifier)
                 .map(member -> getModesFor(member, agent)).ifPresent(memberModes -> {
                     if (!memberModes.contains(ACL.Write)) {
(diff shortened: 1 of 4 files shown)


components/webac/src/main/java/org/trellisldp/webac/WebACService.java
```diff
@@ -273,6 +273,18 @@ private static IRI cleanIdentifier(final IRI identifier) {
         return rdf.createIRI(cleanIdentifier(identifier.getIRIString()));
     }
 
+    /**
+     * Get the structural-logical container for this resource.
+     *
+     * @param identifier the resource identifier
+     * @return a container, if one exists. Only the root resource would return empty here.
+     */
+    private static Optional<IRI> getContainer(final IRI identifier) {
```
ajs6f: This looks like the default impl of `getContainer`, no? If so, should we factor this out somewhere outside of WebAC for potential use elsewhere, or am I being paranoid...?

acoburn: It is _exactly_ the default impl of `getContainer`. One possibility is to add this as a method in `org.trellisldp.api.RDFUtils`. In fact, if you like that idea, I'll do just that.

ajs6f: Hm. I was thinking that `RDFUtils` is more generic, more like stuff usable on _any_ RDF, but something called `BlahUtils` pretty much by name alone does not have very precise semantics. 😁 

Would you be amenable to introducing a `LDPUtils` sister class to `RDFUtils`? Or is that more annoying because people will have to find it...

acoburn: I agree that `RDFUtils` is something of a misnomer, and not just because of this method. I was thinking of renaming it to `TrellisUtils`.

ajs6f: That would be better than introducing a new `BlahUtils`. For the mo, moving the method there is good enough for me!

acoburn:
Yes, exactly. This will make it really easy to rework the `ResourceService::getContainer`.

PR #270, opened 2018-11-02
Rework MementoService interface

### LINKED ISSUE #263: Evaluate methods in MementoService interface
The `MementoService` contains the following methods that need to be implemented by a persistence layer:

```java
CompletableFuture<Void> put(IRI identifier, Instant time, Stream<Quad> data);
 
CompletableFuture<Void> delete(IRI identifier, Instant time);

CompletableFuture<Resource> get(IRI identifier, Instant time);

CompletableFuture<List<Instant>> list(IRI identifier);
```

This issue is a placeholder to evaluate the design of this interface. For example, would it be better for `list` to return a `Stream<Instant>` (and be renamed)? Also, the `delete` method isn't currently used anywhere in the code: should it be removed until such time that it would really be used? Or would it be useful to keep it there for now? (removing it reduces the burden of implementation somewhat)

acoburn:
Resolves #263

This removes the dependency on `commons-lang3` and changes the `MementoService::list` method to `MementoService::mementos`, now returning a `CompletableFuture<SortedSet<Instant>>`

### DIFF
--- a/components/file/src/main/java/org/trellisldp/file/FileBinaryService.java
+++ b/components/file/src/main/java/org/trellisldp/file/FileBinaryService.java
@@ -33,7 +33,6 @@ import static org.apache.commons.codec.digest.MessageDigestAlgorithms.SHA_1;
 import static org.apache.commons.codec.digest.MessageDigestAlgorithms.SHA_256;
 import static org.apache.commons.codec.digest.MessageDigestAlgorithms.SHA_384;
 import static org.apache.commons.codec.digest.MessageDigestAlgorithms.SHA_512;
-import static org.apache.commons.lang3.StringUtils.stripStart;
 import static org.slf4j.LoggerFactory.getLogger;
 import static org.trellisldp.api.TrellisUtils.findFirst;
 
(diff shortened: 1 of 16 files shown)


components/file/src/main/java/org/trellisldp/file/FileBinaryService.java
```diff
@@ -228,7 +227,7 @@ public String generateIdentifier() {
     private File getFileFromIdentifier(final IRI identifier) {
         requireNonNull(identifier, "Identifier may not be null!");
         return of(identifier).map(IRI::getIRIString).filter(x -> x.startsWith("file:")).map(URI::create)
-            .map(URI::getSchemeSpecificPart).map(x -> stripStart(x, "/")).map(x -> new File(basePath, x))
+            .map(URI::getSchemeSpecificPart).map(x -> trimStart(x, "/")).map(x -> new File(basePath, x))
```
ajs6f: I never noticed this-- does it work on Windows?

acoburn: Good question. It is certainly being tested on Windows, so my assumption has been that it works there.

PR #275, opened 2018-11-04
Correct ixn models for special resources, one CDI annotation

ajs6f:


ajs6f:
I don't know if that was the right thing to do for Gradle. Honestly, I wonder if we should just merge `-vocabulary` into `-api`.

ajs6f:
This isn't going to build until `http://www.trellisldp.org/ns/trellis#` is updated...

ajs6f:
Hang on, I just realized that this makes no sense. Never mind.

Issue #245, opened 2018-10-15
Memento and Binaries


ajs6f:
I see methods in `MementoService` for managing `Resource`s, but not `Binary`s. Is the mechanism that a fresh `Binary` is created for every new Memento of a NonRDFSource, and the connections in time are made among the `Resource`s that reference those `Binary`s?

acoburn:
Basically yes. Each `NonRDFSource` has a unique identifier that is handled by the underlying `BinaryService`. If one were to `PUT` a new binary resource, a new identifier would be created for that binary, and so the Memento for the first version would point to the first binary identifier and the second version would point to the second binary.

ajs6f:
Cool, that makes sense.

ajs6f:
Further question: we offer a `MementoService::delete` method. Do we need to write down our assumptions about GC? If the last resource (a Memento) connected to a `Binary` is deleted, should the backend be permitted to delete the `Binary`?

acoburn:
This is precisely what makes `::delete` difficult, whether it's in the `MementoService` or the `ResourceService`: knowing when a binary resource can be deleted. At present, this issue is sidestepped by just _not_ allowing clients to delete Mementos, hence Binaries are never deleted. In fact, there used to be a `ResourceService::purge` [method](https://trellis-ldp.github.io/trellis/apidocs/org/trellisldp/api/ResourceService.html#purge-org.apache.commons.rdf.api.IRI-) that returned a `Stream<IRI>`, containing the identifiers of Binary resources that could be safely deleted. That method has been removed from `master` in part because it was not tied to the HTTP layer in any way but also partly because the entire design of _resource deletion_ needs further examination.

For the 0.8 release, I am somewhat inclined to just _remove_ all the delete-related methods. I would like to see them restored in a subsequent release (0.9?) but only after we figure out exactly what `::delete` really means.

Part of the challenge has to do with the fact that binary resources are loosely coupled to the other services by means of an IRI. That loose coupling hash a lot of advantages, but it also means that a given IRI may be referenced by the MementoService, the ResourceService, both or neither. There is also no efficient mechanism to determine which resources/mementos refer to a given Binary resource.

ajs6f:
> For the 0.8 release, I am somewhat inclined to just remove all the delete-related methods.

This.

ajs6f:
[...]

PR #276, opened 2018-11-04
Pruning MementoService

ajs6f:
Addresses discussion in:

https://github.com/trellis-ldp/trellis/issues/263
https://github.com/trellis-ldp/trellis/issues/245

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/MementoService.java
+++ b/core/api/src/main/java/org/trellisldp/api/MementoService.java
@@ -16,10 +16,8 @@ package org.trellisldp.api;
 import java.time.Instant;
 import java.util.SortedSet;
 import java.util.concurrent.CompletableFuture;
-import java.util.stream.Stream;
 
 import org.apache.commons.rdf.api.IRI;
-import org.apache.commons.rdf.api.Quad;
 
 /**
  * An interface for a Memento subsystem. Mementos of {@link Resource}s may be made and retrieved using this service.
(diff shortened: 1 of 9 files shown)


core/api/src/main/java/org/trellisldp/api/MementoService.java
```diff
@@ -30,29 +28,16 @@
...
-     * successfully created in the corresponding persistence layer. In the case of an unsuccessful write operation,
-     * the {@link CompletableFuture} will complete exceptionally and can be handled with
-     * {@link CompletableFuture#handle}, {@link CompletableFuture#exceptionally} or similar methods.
-     */
-    CompletableFuture<Void> put(IRI identifier, Instant time, Stream<? extends Quad> data);
-
     /**
      * Create a new Memento for a resource.
      * @param resource the resource
+     * @param time the time at which this Memento should be created
```
ajs6f: This isn't really very good-- what do we want to say here? The time at which the Memento is _defined_?

ajs6f:
I ended up leaving a `::put` signature of 
```
CompletableFuture<Void> put(Resource resource, Instant time)
```
but when I look around Trellis, outside of test code, I only see uses wherein the time to be set is take directly from the resource, specifically [here](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/impl/MutatingLdpHandler.java#L129) in `-http`.

Should we just cut all the way to:
```
CompletableFuture<Void> put(Resource resource)
```
? It works for me because in `-cassandra` I work with time series and I have all the previous versions in the stream anyway. If I'm always saving with the original timestamp, I don't really need to save anything. (Which would be quite rad.)

ajs6f:
See https://github.com/trellis-ldp/trellis/pull/277 for an additional commit with the further pruning described above:
```
CompletableFuture<Void> put(Resource resource, Instant time)
```
=>
```
CompletableFuture<Void> put(Resource resource)
```

PR #277, opened 2018-11-04
Pruning MementoService

ajs6f:
Follow on to:

https://github.com/trellis-ldp/trellis/pull/276

with additional commit for:
```
CompletableFuture<Void> put(Resource resource, Instant time)
```
=>
```
CompletableFuture<Void> put(Resource resource)
```


ajs6f:
Kill 'em all, let the compiler sort 'em out.

PR #286, opened 2018-11-07
Allow Dropwizard apps to use ENV properties

acoburn:
See related issue in the trellis-ext-db repository: https://github.com/trellis-ldp/trellis-ext-db/issues/9

### DIFF
--- a/platform/server/src/docker/config.yml
+++ b/platform/server/src/docker/config.yml
@@ -31,52 +31,54 @@ logging:
     io.dropwizard: INFO
 
 # This may refer to a remote Triplestore, e.g. https://example.org/sparql
-resources: /opt/trellis/data/rdf
+resources: ${TRELLIS_RESOURCES:-/opt/trellis/data/rdf}
 
-binaries: /opt/trellis/data/binaries
+binaries: ${TRELLIS_BINARIES:-/opt/trellis/data/binaries}
 
-mementos: /opt/trellis/data/mementos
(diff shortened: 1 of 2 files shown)


platform/server/src/docker/config.yml
```diff
@@ -31,52 +31,54 @@ logging:
     io.dropwizard: INFO
 
 # This may refer to a remote Triplestore, e.g. https://example.org/sparql
-resources: /opt/trellis/data/rdf
+resources: ${TRELLIS_RESOURCES:-/opt/trellis/data/rdf}
```
atz: It seems like using relative paths in the defaults here would prevent a user having to update many values (or supply many ENV vars).

acoburn: I tend to prefer absolute paths in configuration files because it eliminates the ambiguity of the execution context.

atz: That ambiguity is only avoided in case where everything is installed in the same place anyway.  My premise is that an alternate location (not `/opt/trellis`) is used.  In that case, the clarity is a burdensome incorrectness.

atz: Admittedly, my perspective is coming from the container case, where the execution is highly defined and constrained.  It would be different for an app that might be started up different ways by an admin on CLI.

ajs6f: The containerized use case is certainly very important, but Trellis is fundamentally a framework, not a packaged, ready-to-run application.

Issue #288, opened 2018-11-07
Binary modified vs. resource modified


ajs6f:
Checking through some of the binary-related test code, I have come to realize something that wasn't clear to me (I think).

Am I correct to assume that for a given LDP-NR, the modification date of the binary and the modification date of the description must be maintained separately by the backend? The tests seem to be comparing ETags and expecting that to be the case.

TBC, that makes sense with what I know of ETags, I just want to confirm it here, because by gum, there is a lot of confusion over the relationship between things and their descriptions in LDP, and we don't want to add to that.

ajs6f:
Jsut to be a little more clear, this is what I understand to be indicated by the difference between `Resource::getModified` and `someResource.getBinary()::getModified`.

acoburn:
Yes, those are two distinct values. If a description is updated, then the `Binary::getModified` value wouldn't change (though the `Resource::getModified` would). If the binary itself is updated (e.g. via `PUT`, then -- IIRC -- both values get updated, though you shouldn't expect that those values would be the same).

ajs6f:
Okay, I now suspect that this issue may be related to https://github.com/trellis-ldp/trellis/issues/287, via the following question: for the `MementoService::mementos` method, _which of the two possible mod-dates is to be reported_? Is it a case of "use the main date for RDFSources, but the binary date for NonRDFSources"?

acoburn:
This needs to be confirmed, but if I'm not mistaken, a new (RDF-based) Memento is written whenever a binary or its description is changed. So, the memento service should be using the main date for RDFSources.

acoburn:
I think there may be an issue lurking here. Basically, the DateTime value of the Binary is going to be different than the DateTime value of the RDFSource (upon creation). And if that value varies by more than a second, the values of the TimeMap and/or Memento headers may be misleading.

The issue is also one of design: for Binaries, the modified datetime is set in the [HTTP layer](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/impl/PostHandler.java#L185), while the resource/description is set in the [persistence implementation](https://github.com/trellis-ldp/trellis/blob/master/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java#L167)

[...]

PR #290, opened 2018-11-09
Use template interface with ResourceService

acoburn:
Resovles #288

This introduces the `BinaryTemplate` class and the `ResourceTemplate` interface, which are used for resource creation and replacement.

An implementation of the `ResourceTemplate` interface is available in the `org.trellisldp.http.impl` package and also (for testing) in the `org.trellisldp.test` package.

Most of the changes here are in test code, which is a result of changing the `ResourceService::create` and `::replace` signatures.

The `ResourceTemplate` interface is very similar to the `Resource` interface with two differences:

1. The `ResourceTemplate` does not contain the `::hasAcl` method.
2. The `ResourceTemplate` does not contain the alternate `::stream` methods (i.e. those that accept one or more named graphs and respond with `Triple`s -- just the `Stream<Quad> stream()` method is present, and it should be understood that that method only returns the user-provided quads: no containment, membership or audit quads).

The `BinaryTemplate` is similar to the `Binary` class, except that the `BinaryTemplate` includes `Optional<Instant> getModified()` and a helper method for converting an existing `Binary` object into a `BinaryTemplate`. The reason for the `::getModified` method here is that in the case of a binary description being modified, the metadata for the original binary needs to stay the same, and this allows the HTTP layer to pass along that information to the `::replace` method.

A persistence implementation should take care (in a `::replace` method) to [check for the presence of a modified date](https://github.com/trellis-ldp/trellis/pull/290/files#diff-0f7a35d0461f47ab4b396fe19347264cR192) on the binary and keep it unchanged, if it is available. The HTTP layer will _not_ pass in a datetime value on `::create` operations, and so a persistence implementation should use whatever value is determined appropriate for the resource as a whole.

### DIFF
--- a/components/test/src/main/java/org/trellisldp/test/SimpleResourceTemplate.java
+++ b/components/test/src/main/java/org/trellisldp/test/SimpleResourceTemplate.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.trellisldp.test;
+
+import static java.util.Objects.requireNonNull;
+import static java.util.Optional.of;
+import static java.util.Optional.ofNullable;
+
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.apache.commons.rdf.api.Dataset;
+import org.apache.commons.rdf.api.IRI;
+import org.apache.commons.rdf.api.Quad;
+import org.trellisldp.api.BinaryTemplate;
+import org.trellisldp.api.ResourceTemplate;
+import org.trellisldp.vocabulary.LDP;
+import org.trellisldp.vocabulary.OA;
+import org.trellisldp.vocabulary.Trellis;
+
+/**
+ * A template for generating a {@link org.trellisldp.api.Resource} in a persistence layer.
+ */
+public class SimpleResourceTemplate implements ResourceTemplate {
+
+    private final IRI identifier;
+    private final IRI interactionModel;
+    private final IRI container;
+    private final Dataset dataset;
+    private final BinaryTemplate binary;
+
+    /**
+     * Create a template for generating a Trellis Resource.
+     * @param identifier the identifier
+     * @param interactionModel the LDP interaction model
+     * @param dataset the dataset
+     * @param container the container, may be {@code null}
+     * @param binary the binary template, may be {@code null}
+     */
+    public SimpleResourceTemplate(final IRI identifier, final IRI interactionModel, final Dataset dataset,
+            final IRI container, final BinaryTemplate binary) {
+        requireNonNull(identifier, "identifier may not be null!");
+        requireNonNull(interactionModel, "interactionModel may not be null!");
+        this.identifier = identifier;
+        this.interactionModel = interactionModel;
+        this.container = container;
+        this.dataset = dataset;
+        this.binary = binary;
+    }
+
+    @Override
+    public IRI getIdentifier() {
+        return identifier;
+    }
+
+    @Override
+    public IRI getInteractionModel() {
+        return interactionModel;
+    }
+
+    @Override
+    public Optional<IRI> getMembershipResource() {
+        return firstValue(LDP.membershipResource);
+    }
+
+    @Override
+    public Optional<IRI> getMemberOfRelation() {
+        return firstValue(LDP.isMemberOfRelation);
+    }
+
+    @Override
+    public Optional<IRI> getMemberRelation() {
+        return firstValue(LDP.hasMemberRelation);
+    }
+
+    @Override
+    public Optional<IRI> getInsertedContentRelation() {
+        return firstValue(LDP.insertedContentRelation);
+    }
+
+    @Override
+    public Optional<IRI> getContainer() {
+        return ofNullable(container);
+    }
+
+    @Override
+    public Optional<BinaryTemplate> getBinaryTemplate() {
+        return ofNullable(binary);
+    }
+
+    @Override
+    public Dataset dataset() {
+        return dataset;
+    }
+
+    @Override
+    public Stream<? extends Quad> stream() {
+        return dataset.stream();
+    }
+
+    @Override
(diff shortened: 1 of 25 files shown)


acoburn:
There's an interesting design question here that isn't addressed in the commits thus far: should `MutableDataService::delete` also accept a type `<V>` to make the interface consistent? And should the `ImmutableDataService<U>` have a generic type `<V>` added, resulting in an `ImmutableDataService::add(V data)` method?

I am leaning toward an affirmative answer in both cases, but are there other thoughts? The current structure is pragmatic and effective, but a little less consistent design-wise.

Also, is the `ResourceTemplate::stream` really necessary? One can always get that data by calling `ResourceTemplate.dataset().stream()`

components/test/src/main/java/org/trellisldp/test/SimpleResourceTemplate.java
```diff
@@ -0,0 +1,128 @@
...
+        return dataset;
+    }
+
+    @Override
+    public Stream<? extends Quad> stream() {
+        return dataset.stream();
+    }
+
+    @Override
+    public Stream<Entry<String, String>> getExtraLinkRelations() {
```
ajs6f: If we're going to leave a computation here, we ought to explain in a Javadoc why  `OA.annotationService`and `LDP.inbox` merit this treatment, no?

core/api/src/main/java/org/trellisldp/api/BinaryTemplate.java
```diff
@@ -0,0 +1,105 @@
...
+        return ofNullable(modified);
+    }
+
+    /**
+     * Create a Binary template from existing binary metadata.
+     *
+     * @param binary the binary metadata
+     * @return a new binary template containing the original Binary modification date
+     */
+    public static BinaryTemplate fromBinary(final Binary binary) {
```
ajs6f: Does this belong as an instance method on `Binary` (`toTemplate`)?

core/api/src/main/java/org/trellisldp/api/MutableDataService.java
```diff
@@ -16,50 +16,41 @@
...
 
 /**
  * A service that persists resources by <i>replacing</i> their records.
  *
  * @author ajs6f
  * @param <U> the type of resource that can be persisted by this service
+ * @param <V> the type of data used as a template for generating a resource
  */
-public interface MutableDataService<U> extends RetrievalService<U> {
+public interface MutableDataService<U, V> extends RetrievalService<U> {
```
ajs6f: I'm a bit uncomfortable that we have these two params that are clearly attached to each other semantically, but not in the type system at all. (Nothing prevents me from declaring a `MutableDataService<Resource, BinaryTemplate>`.)

Maybe we could do `Resource < ResourceTemplate` and the same for binaries? Or we could even think about a builder pattern...

ajs6f:
Ok, I'm starting to think that we're missing some kind of insight here. We have a situation in which two abstractions (resource and binary) both have the same structure: more-or-less simple metadata, accompanying a payload. In one case the payload is RDF, in the other it is bits. 

That metadata divides into that which we can derive from an initial request and that which must (or at least, we think _should_) come from the backend.

Lastly, we have a notion of management we want to express-- is the data in the server's remit? That should be visible in the type system.

This all seems to be adding up to a problem in composition, and we're trying to solve it by generating independent, formally unconnected types and using them in different slots (params, return values).

Again, I would be fine seeing this merged, but I think it's making visible that we aren't yet at a 1.0 SPI. I 'm going to spend some time thinking about this today and see if I can offer anything practical we can use.

ajs6f:
Okay, here's a thought-- what if we went to the following signature for the `MutableDataService` methods:
```
create(Metadata m, Stream<Quad> q, InputStream b);
```
with either or both of the last two nullable? All of the accessors from `Resource` and `Binary` would be in the `Metadata` class, some presented as `Optional` return values. So the backend accepts the whole box of metadata, uses what it should, ignores what it shouldn't (or it's just not there, returning `empty()`), and does its thing. The return value of `get` is still a `Resource`. 

Although I still feel weird about the use of `Binary`. Maybe we kill `Binary`, put its metadata methods on `Resource` wrapped in `Optional` if need be, and add also to `Resource` an `Optional<InputStream> getNonRdfContent()` or something like that.

Do you see what I'm getting at, tho'? I think we can profitably separate the metadata (which for both `Resource` and `Binary` is just a box of scalar values, which we can make flexible just by using `Optional`, and that's a legit use case for that type as I understand it) from the content (which is fundamentally different).

Alternatively, if you don't want to pour a glass of`Optional` all over everything, we could do something like`ResourceTemplate > Resource` where we just drop the problematic datetimes that brought us here, and still use
```
create(ResourceTemplate m, Stream<Quad> q, InputStream b);
```
and `Resource` just `extends ResourceTemplate` by adding the relevant datetime info and either or both content types. (And anything else we think that only the backend could know, like `hasAcl`.)

The idea is that clients (i.e. `trellis-http`) come to the API with metadata in one hand and content in the other, because clients can't have a `Resource` unless they got it from `ResourceService`. But anyone can have metadata just by putting values in a box, and anyone can have content, which could come from anywhere. If you bring those two things together and give them to the API together (and they make sense together) you can get a minted `Resource` back.

Does this sound like a direction that could make things clearer? I can send a PR if that would help.

acoburn:
I see two general approaches for this. One is to return to a structure we had earlier, where the (e.g.) `MutableDataService` is defined like:

```java
interface MutableDataService<U> extends RetrievalService<U> {
    CompletableFuture<Void> create(U resource);
    CompletableFuture<Void> replace(U resource);
    CompletableFuture<Void> delete(U resource);
}
```

This results in a symmetry with mutation and retrieval: both interact with a single type. In the context of the `ResourceService`, that type is `Resource`. In this case, the `http` layer would have its own implementation of `Resource` that is passed along to the concrete `ResourceService` (persistence-layer) implementation.

I like the simplicity of this approach. The downside of this is that I don't have a lot of clarity about the `::getModified` methods -- do those return `null` in the case of a not-yet-persisted resource? Or should those be changed to `Optional<Instant>`? That is, incidentally, the crux of the issue that prompted all of this.

The other approach, which you are suggesting above is to partition the arguments into "metadata" and "content" -- though the `InputStream` argument wouldn't go to the `ResourceService`, it would be sent to the `BinaryService`, which may be a completely different technology stack from the `ResourceService` impl. What this approach recognizes is that the content of a resource or binary and the metadata of that resource or binary are really separate things and can be managed separately. We are already managing binary metadata in the `ResourceService`. This line of reasoning could lead us in the direction of another `*Service`, allowing for the separation of "metadata" and "content" of a `Resource`. The downside (from my initial thoughts on this) of this approach is that there is an inextricable relationship between content and metadata: for example: any DC or IC `ldp:*` properties co-exist as RDF content and metadata (hence the `::getMembershipResource` method).

If I had to pick one approach right now, I'd lean toward the former.

Either way, I don't think we should merge this PR in its current state.

ajs6f:
Opening a [ticket](https://github.com/trellis-ldp/trellis/issues/292) to avoid putting too much general discussion on a PR.

acoburn:
Closing this, as the clear consensus is to take a different approach.

PR #293, opened 2018-11-10
Remove wildcard types

### LINKED ISSUE #291: Wildcard types in API
The core API has a number of methods that return a wildcard type:

```java
interface Resource {
    Stream<? extends Quad> stream();
    Stream<? extends Triple> stream(IRI graph);   
}

interface ResourceService {
    CompletableFuture<? extends Resource> get(IRI identifier);
}
```

Do we really need this level of flexibility?

acoburn:
Resolves #291


ajs6f:
I'd like to try building `trellis-cassandra` against this, but I'm not likely to get done with that until tomorrow. Cool?

acoburn:
ping @ajs6f. Just wanted to check whether you still plan to test this on trellis-cassandra.

ajs6f:
Yep, this morning. I was so caught up with our adventures in Memento that I didn't want to be juggling too many PRs at once.

ajs6f:
I'm not in favor of this. By cutting  return values from`CompletableFuture<? extends Resource>` to `CompletableFuture<Resource>`, this PR makes it impossible to implement `ResourceService` with `CompletableFuture<MySpecificTypeOfResource>` which means that methods on `MySpecificTypeOfResource` can't be used out of `get`. E.g. no
```
get(someId).thenApply(MySpecificTypeOfResource::myExtraUsefulMethod)
```

What is the benefit of this PR?

ajs6f:
Another way to put it: if we're going to drop the wildcard (I have no problem with that) let's do it in the most flexible way-- by parameterizing `ResourceService` with a type for the subtype of `Resource` involved. That's flexible, typesafe, and not significantly more burdensome or verbose than "downcasting in advance" to `Resource` itself.

ajs6f:
Would you like a PR with my suggestion above (parameterize the service types)?

acoburn:
My main interest here is removing the wildcard types. Adding extra parameterization would be fine. The `ResourceService` already parameterizes the `RetrievalService` as: `CompletableFuture<T> get(IRI identifier)`. (currently, it's `CompletableFuture<? extends T>`; this PR removes the wildcard). Are there parts of the code where you need that additional flexibility? The HTTP layer doesn't.

That said, we can leave in this flexibility for `::get` (or parameterize it further); my main interest here is changing `Stream<? extends Quad> stream()` to `Stream<Quad> stream()` in the `Resource` interface.

How about if I limit this PR to just that change (the `::stream` method) and we address the other issue separately.

ajs6f:
Sure, I have no use for `<? extends Quad>` right now, and we can open it back up later. I would happy to merge that and we can take a little time and discuss the other types in a ticket.

Issue #287, opened 2018-11-07
Redirections from Memento?


ajs6f:
I'm getting some very queer failures from `MementoResourceIT`. I'm really only concerned here with `testMementoAcceptDateTimeHeader` and `testMementoAcceptDateTimeHeader`, because I'm not at all sure how my backend could be triggering the emission of redirects. (The others I'm chalking up for now to plain old errors in my backend.)

@acoburn, any thoughts? Is there some condition under which a backend can return a resource that gets translated into a redirect by HTTP? I can't imagine why that would happen... let me know if there's any context or additional info that might clear this up.

```
[ERROR] Failures:
[ERROR]   LdpBinaryIT.testPatchBinaryDescription:117 Check that the ETag values are the same ==> expected: <"2b31f4da4305d4825fd8ad5281acfa61"> but was: <"30b217e7d59b2c13b25e845c380a6de9">
[ERROR]   MementoBinaryIT.testMementoAcceptDateTimeHeader Check for a successful memento request ==> expected: <SUCCESSFUL> but was: <REDIRECTION>
[ERROR]   MementoBinaryIT.testMementoDateTimeHeader Check that the memento-datetime header is correct ==> expected: <2018-11-07T16:26Z> but was: <2018-11-07T16:25:59Z>
[ERROR]   MementoResourceIT.testMementoAcceptDateTimeHeader Check for a successful memento request ==> expected: <SUCCESSFUL> but was: <REDIRECTION>
[ERROR]   MementoResourceIT.testMementoDateTimeHeader Check that the memento-datetime header is correct ==> expected: <2018-11-07T16:26:06Z> but was: <2018-11-07T16:26:05Z>
```

ajs6f:
Just to give a slightly clearer report, I fixed up some of those tests based on the understanding I hope to confirm in https://github.com/trellis-ldp/trellis/issues/288 to get what I show below, but I'm still a bit lost on the redirections...
```
[ERROR] Failures:
[ERROR]   MementoBinaryIT.testMementoAcceptDateTimeHeader Check for a successful memento request ==> expected: <SUCCESSFUL> but was: <REDIRECTION>
[ERROR]   MementoBinaryIT.testMementoDateTimeHeader Check that the memento-datetime header is correct ==> expected: <2018-11-07T17:00:58Z> but was: <2018-11-07T17:00:57Z>
[...]

PR #296, opened 2018-11-14
Truncate Instants to second precision

acoburn:
Related to #287 

IIRC, this weird subtraction/addition was in place to support millisecond precision Mementos in certain circumstances. I am not convinced that it is correct, and so this commit just truncates the memento to the closest second, which is the smallest resolution supported by RFC 1123 (upon which Memento is based).

If this fixes the issue referred to, then we should go further and make the `?version=...` URL parameters (for Mementos) have second precision, too. (They currently have millisecond precision)

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/impl/MementoResource.java
+++ b/core/http/src/main/java/org/trellisldp/http/impl/MementoResource.java
@@ -17,6 +17,7 @@ import static java.lang.String.join;
 import static java.time.ZoneOffset.UTC;
 import static java.time.ZonedDateTime.ofInstant;
 import static java.time.format.DateTimeFormatter.RFC_1123_DATE_TIME;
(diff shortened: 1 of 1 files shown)


ajs6f:
Going to test right now...

ajs6f:
Definitely fixing something. Now I'm left with 
```
[ERROR] Failures:
[ERROR]   MementoBinaryIT.testMementoAcceptDateTimeHeader Check for a successful memento request ==> expected: <SUCCESSFUL> but was: <REDIRECTION>
[ERROR]   MementoResourceIT.testMementoAcceptDateTimeHeader:24->lambda$testMementoAcceptDateTimeHeader$1:26 Check for a successful memento request ==> expected: <SUCCESSFUL> but was: <REDIRECTION>
```
so the funky datetime comparison problem is gone!

PR #297, opened 2018-11-14
Fetch LDP-NR descriptions for tests

### LINKED ISSUE #283: LdpBinaryTests  and NonRDFSource descriptions
I'm a bit confused about our handling of NonRdfSource descriptions and how they are being tested. In the [docs](https://github.com/trellis-ldp/trellis/wiki/Resource-Management#retrieving-resources) I see:

> When Non-RDF (binary) resources are added, a corresponding RDF resource is created. Link headers will lead a client to that resource. 

which jibes with my understanding of what we are about; it shouldn't be possible to make a single URI both an RDfSource and a NonRDFSource, so we link out the description (standoff description).

But to my examination, the tests in `LdpBinaryTests` (particularly `testGetBinaryDescription` and `testPatchBinaryDescription`) try to retrieve RDF directly from a NonRDFSource-- no `Link` header is used to intermediate.

This doesn't work for the current `trellis-cassandra`, although I probably could make it work, but is it right? How are those tests passing for other impls?

acoburn:
Resolves #283 

This updates the test code to explicitly fetch LDP-NR descriptions when RDF is being requested.

### DIFF
--- a/components/test/src/main/java/org/trellisldp/test/CommonTests.java
+++ b/components/test/src/main/java/org/trellisldp/test/CommonTests.java
@@ -23,11 +23,13 @@ import static org.trellisldp.api.TrellisUtils.getInstance;
 import static org.trellisldp.test.TestUtils.getLinks;
 import static org.trellisldp.test.TestUtils.hasType;
 
+import java.net.URI;
 import java.util.stream.Stream;
 
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.EntityTag;
+import javax.ws.rs.core.Link;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
@@ -74,6 +76,18 @@ public interface CommonTests {
      */
     default WebTarget target(final String url) {
         return getClient().target(url);
+    }
+
+    /**
+     * Get the describedby Link value, if one exists.
+     * @param url the URL
+     * @return the location of a description resource, or null if none is available
+     */
+    default String getDescription(final String url) {
+        try (final Response res = target(url).request().head()) {
+            return getLinks(res).stream().filter(link -> "describedby".equals(link.getRel()))
+                .map(Link::getUri).map(URI::toString).findFirst().orElse(null);
+        }
     }
 
     /**
--- a/components/test/src/main/java/org/trellisldp/test/LdpBinaryTests.java
+++ b/components/test/src/main/java/org/trellisldp/test/LdpBinaryTests.java
@@ -19,6 +19,7 @@ import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN_TYPE;
 import static org.apache.commons.rdf.api.RDFSyntax.TURTLE;
 import static org.awaitility.Awaitility.await;
+import static org.awaitility.Duration.TWO_SECONDS;
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
@@ -173,7 +174,8 @@ public interface LdpBinaryTests extends CommonTests {
             descriptionETag = res.getEntityTag();
             assertTrue(descriptionETag.isWeak(), "Check for a weak ETag");
         }
-
+        // wait for enough time so that the ETags will surely be different
+        await().pollDelay(TWO_SECONDS).until(() -> true);
         // Patch the description
         try (final Response res = target(descriptionLocation).request().method("PATCH",
                     entity("INSERT { <> <http://purl.org/dc/terms/title> \"Title\" } WHERE {}",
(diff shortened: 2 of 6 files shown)


ajs6f:
Testing now...

ajs6f:
No change to the redirect behavior... this gets weirder. I'll start putting in some logging to see what's going on with this. Why are redirects occurring at all?

ajs6f:
No slashes-- it's a Memento thing. The initial request goes to (example) `http://localhost:63697/95e925e0-605e-4e2c-87b2-d877ceaf4736/a2a3110a-7578-44c0-a9f9-a40574407c15` and the redirect has `Location : [http://localhost:63697/95e925e0-605e-4e2c-87b2-d877ceaf4736/a2a3110a-7578-44c0-a9f9-a40574407c15?version=1542224275000]`.

But this is making some sense. The method in question is in `MementoResourceTests`, which you do not alter in this PR (currently). The code is:
```
final Response res = target(getResourceLocation()).request().header(ACCEPT_DATETIME, date).get()
```
which is unlike the ones you've altered in this PR to use `HEAD` (and following the links). Shouldn't we be using that everywhere? Otherwise it seems like we could be running into variable behavior in JAX-RS clients or who knows what!

ajs6f:
Testing my theory now...

ajs6f:
Yyyyyup. That did it. I'll send a PR to this PR.

ajs6f:
Never mind. This isn't working at all, and I have no idea why. I'm getting intermittent `CLIENT ERROR`s instead of redirects, the exact same  behavior where the client sends one datetime in `Accept-Datetime` and the server responds with an earlier one, everything we thought we fixed.

I'll send a PR so that you can see what I did to try to fix the redirect problem (which _does_ seem to be gone).

I'm heading out for the day, but I'll be back on this hard tomorrow. I want it done, and I'm sure you're tired of hearing about it.

ajs6f:
As a note for the future here, I am getting intermittent 406s, it turns out, because that's what the server does when a client asks for a Memento with a datetime that isn't available. Why _that_ is happening is an interesting question.

acoburn:
Yes, that is the behavior I've also noticed (intermittently). My suspicion is that it relates to the date-time conversion (RFC 1123 <-> java.time.Instant), because it always seems to be an off-by-one-second issue.

ajs6f:
I'm running down the hypothesis now that it's because https://github.com/trellis-ldp/trellis/pull/296 truncated the values for headers without truncated the values _on the way in_ (to persistence).

So far I'm passing the tests.

This might suck, because it might mean that we would have to ask every backend to do this without being able to fully enforce it. Or maybe I'm being too optimistic that I've got it and too pessimistic about the potential solution.

ajs6f:
Ok, that did it. I've sent one more commit to https://github.com/trellis-ldp/trellis/pull/298 and with it, `trellis-cassandra` passes all the relevant tests. The secret ingredient turned out to be truncating not just the data going into the headers, but the data against which it is being compared, to avoid the off-by-one-second problem, which is secretly a off-by-less-than-one-second problem.

We have some more discussion to have about this, because we're asking for certain behavior at deeper layers than HTTP, but that's a different question. I'm all in favor of merging this _with_ https://github.com/trellis-ldp/trellis/pull/298 ASAP!

acoburn:
It turns out that the crux of the `REDIRECT` vs `SUCCESSFUL` response has to do with the configuration of the JAX-RS client. If the client follows redirects (as is the case with Jersey by default), you'd expect a `SUCCESSFUL` response, but in a non-Jersey context (or if Jersey is configured _not_ to follow redirects), you'd see the `REDIRECT` response. Either way, that's not what that test is trying to check: the test is checking the response headers _after_ redirection. So after I do a little more testing, I will be adding another commit here so that the test code isn't making assumptions about the configuration of the JAX-RS client. That should get the integration tests to pass (they currently don't)

ajs6f:
Right, that's what I meant about clients at the end of [this](https://github.com/trellis-ldp/trellis/pull/297#issuecomment-438793147).

ajs6f:
The other piece of the #287 puzzle is the truncation question. I can reliably pass all the tests now, because I understand that. We need to get that down somehow in the codebase, in a comment if nothing else, but I'd like to see if we can be stronger about it. It's subtle and primed to cause problems.

We should, in all bravery, also take up the granularity question on the Memento-dev list. I think we have some valid use cases for distinguishing Mementos with less than a second's difference between them. But maybe not (from the Memento POV, which _is_ primarily about web archiving) and if so we should clarify that.

ajs6f:
Come to think of it, _can_ we make some assumptions about client behavior? IOW, is there an "autoredirect" switch in the JAX-RS API, or only in impl-specific configs? Probably the latter, but I'll go take a look.

acoburn:
With Jersey, there is a property that can be set on the `Client` object, but it's an impl-specific value.

ajs6f:
Yeah, see my edit above. :(

acoburn:
Travis-CI is now happy with the state of this PR. Any objections to merging it now?

ajs6f:
Hell, no. :sailboat: it.

PR #308, opened 2018-11-19
Use an InputStream for incoming data

### LINKED ISSUE #306: Buffering binaries in the HTTP layer
All incoming RDF content must be buffered in memory in order to parse and verify the incoming data.

NonRDF binaries, however, shouldn't need to be buffered in this way. At present, however, all incoming (`PUT` and `POST`) binary data is buffered in a `File` object before being sent along to the `BinaryService`. In some cases (e.g. verifying digests) it may be necessary to buffer the data to a local file, but in the general case, it should be possible to completely avoid this.

acoburn:
Resolves #306

Incoming user data will no longer be buffered in a local `File`. Instead, bytes will be streamed directly to their destination, whether that's an RDF `Graph` or the backend `BinaryService`.

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/impl/MutatingLdpHandler.java
+++ b/core/http/src/main/java/org/trellisldp/http/impl/MutatingLdpHandler.java
@@ -44,6 +44,7 @@ import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response.ResponseBuilder;
 import javax.ws.rs.core.StreamingOutput;
 
+import org.apache.commons.io.input.CountingInputStream;
 import org.apache.commons.rdf.api.Graph;
 import org.apache.commons.rdf.api.IRI;
 import org.apache.commons.rdf.api.Quad;
@@ -56,6 +57,7 @@ import org.trellisldp.api.Resource;
 import org.trellisldp.api.RuntimeTrellisException;
 import org.trellisldp.api.ServiceBundler;
 import org.trellisldp.api.Session;
+import org.trellisldp.http.core.Digest;
 import org.trellisldp.http.core.TrellisRequest;
 import org.trellisldp.vocabulary.AS;
 import org.trellisldp.vocabulary.LDP;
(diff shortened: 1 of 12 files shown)


acoburn:
I want to do a bit more work on this. I think it should be possible to completely eliminate any File buffering in the HTTP layer. For instance, by extending `BinaryMetadata` to include:

```java
class BinaryMetadata {
    Optional<String> getDigest();
    Optional<String> getAlgorithm();
}
```
those values can be passed along to the `BinaryService` to handle in whatever fashion makes sense for that implementation.

ajs6f:
I'm all in favor of streaming all the way into the backend, for fairly obvious reasons. Lemme know when you want more eyes on this PR.

acoburn:
@ajs6f it would be great if you could look at this. The feature we lose with these commits is the size of binaries. Before, we could just call `File::length` to get that. Now, the best option appears to be wrapping the `InputStream` in a `org.apache.commons.io.input.CountingInputStream` and then calling `::getByteCount` -- unless you know of a better way to do that. Either way, I'd like to keep that change as a separate issue, just to keep the commits simpler.

acoburn:
I added support for keeping track of binary `InputStream` sizes and simplified the `BinaryMetadata` class.

core/http/src/main/java/org/trellisldp/http/impl/MutatingLdpHandler.java
```diff
@@ -98,11 +100,13 @@ protected MutatingLdpHandler(final TrellisRequest req, final ServiceBundler trel
      * @param entity the entity
      */
     protected MutatingLdpHandler(final TrellisRequest req, final ServiceBundler trellis,
-            final String baseUrl, final File entity) {
+            final String baseUrl, final InputStream entity) {
         super(req, trellis, baseUrl);
         this.entity = entity;
         this.session = ofNullable(req.getSecurityContext().getUserPrincipal()).map(Principal::getName)
             .map(getServices().getAgentService()::asAgent).map(HttpSession::new).orElseGet(HttpSession::new);
+        this.algorithms = trellis.getBinaryService().supportedAlgorithms().stream().map(String::toLowerCase)
```
ajs6f: If we're gong to call this on every mutating request, we should memoize it and assume that the supported digest algorithms aren't going to change in between two requests.

acoburn: or, just call this when a digest-based request is used, which is already going to be a computationally expensive operation.

acoburn:
Regarding making the `size` issue the responsibility of the backend, that is an entirely feasible approach (and I rather like that idea). In that case, rather than returning an `InputStream`, perhaps the `BinaryService` would return a `Binary` such as:

```java
class Binary {
   public Binary(InputStream in, Long size) { }
   Long getSize() { }
   InputStream getInputStream() { }
}
```

acoburn:
For the record, I don't like the `PersistenceResult` class either, and I'd be happy to dispense with it.

ajs6f:
Just to be clear, you mean that `BinaryService::getContent` would return that, yes? If so, sounds good, and we should then change the method names to `BinaryService::get` in alignment with `ResourceService`.

You'll also want to add:
```
InputStream getInputStream(int from, int to) ;
```
to keep parity with what we have now.

ajs6f:
Also, if we do that, I think `BinaryService extends RetrievalService<Binary>` which is nice and symmetrical.

acoburn:
This sounds like a very good plan, though I probably won't get to it until next week.

ajs6f:
Cool. I might take a crack at it between now and then, because `trellis-c*` is mostly waiting on Tamaya to straighten out its SNAPSHOT builds.

acoburn:
Great, you should be able to make use of all of the first commit here. It would probably be easiest to throw out the second commit (though we'd still want to remove the `digest` and `algorithm` parts of the `BinaryMetadata` class). Also, the `Binary` mentioned above should very probably be an interface and not a class, since a backend will certainly have implementation-specific ways to handle range requests.

ajs6f:
I'm taking it back to `master` to start with, because I want to work out from the API-- PR in a moment to discuss.

acoburn:
Superseded by #314

PR #310, opened 2018-11-20
Turn down exception logging

### LINKED ISSUE #309: Stacktraces with every HTTP error response
The HTTP layer currently will log _every_ error with a stack trace, based on the line [here](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java#L366).

This includes all 4xx and 5xx errors. And while this is great for debugging, I wonder how useful a stacktrace would be in the logs for every 404 response.

For example:

```
ERROR [2018-11-20 19:25:21,664] org.trellisldp.http.TrellisHttpResource: Error:
! javax.ws.rs.NotFoundException: HTTP 404 Not Found
! at org.trellisldp.http.impl.GetHandler.initialize(GetHandler.java:143)
! at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642)
! ... 8 common frames omitted
! Causing: java.util.concurrent.CompletionException: javax.ws.rs.NotFoundException: HTTP 404 Not Found
! at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
! at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
! at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:645)
! at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
! at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1705)
! at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1692)
! at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
! at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
! at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
! at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
! at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
0:0:0:0:0:0:0:1 - - [20/Nov/2018:19:25:21 +0000] "GET /blahblahblah HTTP/1.1" 404 240 "-" "curl/7.54.0" 25
```

I'd like to suggest changing the log level of that line to `DEBUG`.

acoburn:
Resolves #309

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
+++ b/core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
@@ -363,7 +363,7 @@ public class TrellisHttpResource {
     }
 
     private Response handleException(final Throwable err) {
-        LOGGER.error("Error:", err);
+        LOGGER.debug("Trellis Error:", err);
         return of(err).map(Throwable::getCause).filter(WebApplicationException.class::isInstance)
             .map(WebApplicationException.class::cast).orElseGet(() -> new WebApplicationException(err)).getResponse();
     }


ajs6f:
My problem is this: I don't want `ERROR` logging for every 40x any more than you do, but I _do_ want `ERROR` logging for 50xs. 

Why are 40x hitting this logging statement? They aren't exceptions, are they? They are `MISSING_RESOURCE`.

acoburn:
The issue has to do with how we use the JAX-RS exception handlers. When we throw a `BadRequestException` it allows us to abort the execution of a request. The alternative was to create a special `Response.Builder` for every such 4xx error and pass that along the entire async handler chain, which made the code really cumbersome.

Generally speaking, a 5xx error should be happening in the service implementation, and _that_ is where the stacktrace should be displayed.

ajs6f:
The problem with that is twofold: it means duplicate code in every impl, and it's not possible to know when a `RuntimeException` will arise, so try-catch-log code would have to be _everywhere_, just in case.

I don't understand the problem with logging here-- if we know that `BadRequestException` is a 40x, why don't we just not log `BadRequestException`s (but log other throwables) at this point?

acoburn:
> if we know that BadRequestException is a 40x, why don't we just not log BadRequestExceptions (but log other throwables) at this point?

This is probably the best way to handle this issue

ajs6f:
Ok, I'll send a PR to this PR.

PR #312, opened 2018-11-21
More subtle logging for web exceptions

ajs6f:


ajs6f:
For https://github.com/trellis-ldp/trellis/pull/310

core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
```diff
@@ -363,7 +364,9 @@ private String getBaseUrl(final TrellisRequest req) {
     }
 
     private Response handleException(final Throwable err) {
-        LOGGER.debug("Trellis Error:", err);
+        if (!(err instanceof BadRequestException)) {
```
acoburn: These are typically going to be instances of `CompletionException`; I think that what you'd want here is: `if (!(err.getCause() instanceof BadRequestException)) { ... }`

ajs6f: Fine with me-- the backends don't throw this, of course, so wherever it is being thrown is in the `-http` code and you will know more about the specifics than I could determine at runtime.

PR #314, opened 2018-11-21
Binaries fully streaming into the backend

ajs6f:
NOT FOR MERGING YET 

https://github.com/trellis-ldp/trellis/issues/306

### DIFF
--- a/components/file/src/main/java/org/trellisldp/file/FileBinary.java
+++ b/components/file/src/main/java/org/trellisldp/file/FileBinary.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.trellisldp.file;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+
+import org.apache.commons.io.input.BoundedInputStream;
+import org.slf4j.Logger;
+import org.trellisldp.api.Binary;
+
+/**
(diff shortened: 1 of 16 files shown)


ajs6f:
This PR will not build yet-- I have taken the digest-checking code out of the HTTP layer but not yet added it back into the backends. Question: there are many many tests in the HTTP module that are currently failing as described above. They no longer belong in the HTTP layer-- should they be moved to `trellis-test`? 

Also do we even want digest checking as atomic with ingestion? What about a design like:
```
CompletableFuture<String> result = bS.setContent(metadata, entity)
                   .thenComposeAsync(voidValue->bS.calculateDigest(metadata.getIdentifier(), "someDigestType"));
```

ajs6f:
In the non-atomic design above, the client to `BinaryService` is responsible for deleting the bad bitstream, not the backend. I like that, because it means that behavior is uniform across backends, but it's undeniably less "transactionalish".

acoburn:
W/R/T digest handling above, I like the idea of putting that into a default method, but it would be more efficient to use a DigestInputStream as a filter on the way in, and then compare that value to the expectation. If they don’t match, then issue the purgeContent method

ajs6f:
Well, I thought about that, but I intentionally left the actual digest calculation to the backend in all cases because the backend may have _much_ more efficient ways of doing that. But I'm okay with doing the calc in a `default` method in `BinaryService`, precisely because the impl can override that with its own jazz.

I'll add a commit with that sometime later today.

ajs6f:
> Leaving the client responsible for cleaning up a bad bitstream is the design I also landed on, so let’s continue to pursue that model.

Yeah, and actually, it occurs to me that it might be useful to have the bad bitstream left behind for evidence, anyway.

acoburn:
As part of the #315 PR, I looked at the possibility of putting the digest check into the `BinaryService`, but I ended up not pursing that for the simple reason that a digest mismatch needs to throw a 4xx error and not a 5xx error (which is reserved for "some sort of error in the persistence layer"). And the only way to distinguish between the two is either: introduce some sort of return type (which neither of us liked in #308) or b) add a new Exception type in the Trellis API, which seems a bit overly specific. So in the end, the digest check is still in the HTTP layer. I do like the idea of putting it into the `BinaryService`, but perhaps we can save that for a separate issue.

ajs6f:
Yeah, I'm really not excited about that for the reasons I've already said-- I really want this to be override-able. Maybe we're working in the wrong space, though-- maybe we want a `DigestService` with default impl, such as we've used elsewhere?

acoburn:
I think you're right that the best path will be with some sort of `DigestService`. I'd first like to get the changes in this PR into `master` and then pick up the digest question in a separate issue.

ajs6f:
Yes, it's definitely a step in the right direction. We can open a whole 'nuther issue for the override question.

PR #318, opened 2018-11-26
Open BinaryService return type

ajs6f:
I'd like to be able to return a `CompletableFuture<MyImplResource>` from `get` to reuse inside my impl.


acoburn:
There were some tests that needed to be adjusted. Resolved with e73d5486ee542a8a9a5d479188bf445986e7c153

PR #320, opened 2018-11-27
Move incoming binary digest handling to BinaryService

### LINKED ISSUE #316: Make it possible to override the digest operation for incoming Binaries
For incoming Binaries, the digest computation is handled in the HTTP layer. If this is moved into the BinaryService (perhaps via a new `DigestService` interface), then it would be possible to override a default implementation for a given binary service backend.

Some related discussion can be found at https://github.com/trellis-ldp/trellis/pull/314#issuecomment-440798300

acoburn:
Resolves #316

This moves the processing of incoming binary streams to the
BinaryService by adding an additional BinaryService::setContent method.
This method accepts an additional argument: `algorithm`, and responds
with `CompletableFuture<String>` where that response string is the
value of the computed digest. If the value of that digest is null, one
should take that to mean that the supplied algorithm was not supported.

If the digest value does not match the expected value, it is the
reponsibility of the client code (e.g. in the HTTP layer) to call
BinaryService::purgeContent. A runtime exception, on the other hand,
would indicate a server error.

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/BinaryService.java
+++ b/core/api/src/main/java/org/trellisldp/api/BinaryService.java
@@ -13,10 +13,6 @@
  */
 package org.trellisldp.api;
 
-import static java.util.Base64.getEncoder;
-import static java.util.concurrent.CompletableFuture.completedFuture;
-import static org.slf4j.LoggerFactory.getLogger;
-
 import java.io.InputStream;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
@@ -47,23 +43,19 @@ public interface BinaryService extends RetrievalService<Binary> {
      * Set the content for a binary object using a digest algorithm.
      *
      * @implSpec The default implementation will compute a digest while processing the {@code InputStream}.
-     *           A {@code null} response value indicates that the digest algorithm is not supported. In the
-     *           case of a non-supported algorithm, the content will not be sent along to the binary service.
      * @param metadata the binary metadata
      * @param stream the context
      * @param algorithm the digest algorithm
-     * @return the new completion stage containing the server-computed digest. A {@code null} value
-     *         can be understood to mean that the digest algorithm is unsupported.
+     * @return the new completion stage containing the server-computed digest.
      */
-    default CompletableFuture<String> setContent(final BinaryMetadata metadata, final InputStream stream,
+    default CompletableFuture<byte[]> setContent(final BinaryMetadata metadata, final InputStream stream,
             final String algorithm) {
         try {
             final DigestInputStream input = new DigestInputStream(stream, MessageDigest.getInstance(algorithm));
             return setContent(metadata, input)
-                .thenApply(future -> getEncoder().encodeToString(input.getMessageDigest().digest()));
+                .thenApply(future -> input.getMessageDigest().digest());
         } catch (final NoSuchAlgorithmException ex) {
-            getLogger(BinaryService.class).warn("Unsupported algorithm: {}", ex.getMessage());
-            return completedFuture(null);
+            throw new IllegalArgumentException("Invalid message digest", ex);
         }
     }
 
--- a/core/api/src/main/java/org/trellisldp/api/BinaryService.java
+++ b/core/api/src/main/java/org/trellisldp/api/BinaryService.java
@@ -16,7 +16,6 @@ package org.trellisldp.api;
 import java.io.InputStream;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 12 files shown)


acoburn:
@ajs6f any thoughts on this PR? or shall I go ahead and merge it?

ajs6f:
Sorry @acoburn-- I will take a look and comment later this morning!

core/api/src/main/java/org/trellisldp/api/BinaryService.java
```diff
@@ -36,6 +43,30 @@
      */
     CompletableFuture<Void> setContent(BinaryMetadata metadata, InputStream stream);
 
+    /**
+     * Set the content for a binary object using a digest algorithm.
+     *
+     * @implSpec The default implementation will compute a digest while processing the {@code InputStream}.
+     *           A {@code null} response value indicates that the digest algorithm is not supported. In the
```
ajs6f: I don't like this at all. A unexceptionally-completed `CompletionStage`, to my mind, should _always_ mean that persistence succeeded. Can we go to an exception instead here?

core/api/src/main/java/org/trellisldp/api/BinaryService.java
```diff
@@ -36,6 +43,30 @@
...
+     *           case of a non-supported algorithm, the content will not be sent along to the binary service.
+     * @param metadata the binary metadata
+     * @param stream the context
+     * @param algorithm the digest algorithm
+     * @return the new completion stage containing the server-computed digest. A {@code null} value
+     *         can be understood to mean that the digest algorithm is unsupported.
+     */
+    default CompletableFuture<String> setContent(final BinaryMetadata metadata, final InputStream stream,
```
ajs6f: I like the basic signature here, but are you sure we wouldn't rather this be a `byte[]` rather than a `String`? I ask because `String` implies an encoding, but we don't really have that here. I think that's why `MessageDigest::digest`, etc. return a `byte[]` instead.

acoburn: Using a `byte[]` would be fine. The HTTP layer has a Base64-encoded `String` (which comes from the incoming `Digest` header), but we can just do that conversion in the HTTP layer.

core/api/src/main/java/org/trellisldp/api/BinaryService.java
acoburn: I don't see why we couldn't use an exception here.

core/api/src/main/java/org/trellisldp/api/BinaryService.java
ajs6f: Yeah, that's exactly what I meant.

core/api/src/main/java/org/trellisldp/api/BinaryService.java
ajs6f: Cool!

core/api/src/main/java/org/trellisldp/api/BinaryService.java
acoburn: Thinking about this some more, we should change the signature of the `CompletableFuture<String> calculateDigest(IRI identifier, String algorithm)` to also return `byte[]`

ajs6f: Oh, duh. Yep!

core/api/src/main/java/org/trellisldp/api/BinaryService.java
```diff
@@ -36,6 +39,26 @@
...
+     * @return the new completion stage containing the server-computed digest.
+     */
+    default CompletableFuture<byte[]> setContent(final BinaryMetadata metadata, final InputStream stream,
+            final String algorithm) {
+        try {
+            final DigestInputStream input = new DigestInputStream(stream, MessageDigest.getInstance(algorithm));
+            return setContent(metadata, input)
+                .thenApply(future -> input.getMessageDigest().digest());
+        } catch (final NoSuchAlgorithmException ex) {
+            throw new IllegalArgumentException("Invalid message digest", ex);
```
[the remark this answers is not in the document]
acoburn: Actually, what would be even better would be to change the `String algorithm` argument to `MessageDigest algorithm` (also in the `caclulateDigest` method). This would allow the HTTP layer to handle the `NoSuchAlgorithmException` more easily (i.e. entirely in the HTTP layer). With an `IllegalArgumentException` tucked inside a default method, this seems a bit brittle.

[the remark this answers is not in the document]
ajs6f: Oh, nice. My only concern would be that you can (in theory) reset and reuse a `MessageDigest`, which is why I didn't suggest it (but a `byte[]`) instead of a `String`. I suppose that's a security concern, at some level, but…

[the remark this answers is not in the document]
acoburn: Just to clarify, when I suggested `MessageDigest`, I meant as an argument, not a return value. Thus, the signature would be:

    CompletableFuture<byte[]> calculateDigest(IRI identifier, MessageDigest digest);

and:

    CompletableFuture<byte[]> setContent(BinaryMetadata metadata, InputStream stream,
            MessageDigest digest)

acoburn:
@ajs6f are you happy with the state of this?

core/api/src/main/java/org/trellisldp/api/BinaryService.java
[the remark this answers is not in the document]
ajs6f: Oh, gotcha, that makes sense, and I like it! The exception handling makes much more sense in the HTTP layer.

ajs6f:
LGTM  :rocket:

PR #327, opened 2018-12-10
Move injected contexts out of TrellisRequest class

### LINKED ISSUE #324: Full JAX-RS server-side asynchrony
I've started [a branch of `trellis-cassandra`](https://github.com/ajs6f/trellis-cassandra/tree/Threadpooling) working towards full asynchrony on the backend. (Currently, `t-c*`'s threadpooling simply uses `Runnable::run`, which synchronizes operations much like the other backends). My backend is working fine, but we have a few odd problems in HTTP, and they can only be solved at the Trellis layer.

Specifically, the `*Handler` classes (e.g. `GetHandler`) are using JAX-RS-injected  context like headers, query params, etc. often via `TrellisRequest`. The problem with this is that in fully asynchronous operation, a thread that may be working on a response may not be the thread that originally answered it (which is, after all, the whole point of asynchronous operation). It therefore may not have the same JAX-RS context and operations can crack out with something like e.g. for RESTEasy:
```
RESTEASY003880: Unable to find contextual data of type: org.jboss.resteasy.spi.HttpRequest
```
The suggested correction from [RESTEasy](http://docs.jboss.org/resteasy/docs/3.6.2.Final/userguide/html/Reactive.html#d4e2088) is to set up classes like the `*Handler`s with all the info they will need disconnected from the JAX-RS context, to ensure that any available thread can safely move their work. In our case I suppose that would mean the HTTP headers, some query params, any content stream... this isn't a small amount of work, but it promises great rewards-- the performance improvements from full asynchrony are very very real.

acoburn:
Resolves #324

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
+++ b/core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
@@ -173,16 +173,14 @@ public class TrellisHttpResource {
      *           <a href="https://tools.ietf.org/html/rfc7089#section-4.2.1">section 4.2.1 of RFC 7089</a>.
      * @param response the async response
      * @param uriInfo the URI info
-     * @param secContext the security context
      * @param headers the HTTP headers
      * @param request the request
      */
     @GET
     @Timed
     public void getResource(@Suspended final AsyncResponse response, @Context final Request request,
-            @Context final UriInfo uriInfo, @Context final HttpHeaders headers,
-            @Context final SecurityContext secContext) {
-        fetchResource(new TrellisRequest(request, uriInfo, headers, secContext))
+            @Context final UriInfo uriInfo, @Context final HttpHeaders headers) {
+        fetchResource(new TrellisRequest(request, uriInfo, headers))
             .thenApply(ResponseBuilder::build).exceptionally(this::handleException).thenApply(response::resume);
     }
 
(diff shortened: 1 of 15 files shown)


acoburn:
@ajs6f there is some more clean-up that can be done here, but is this what you had in mind? You'd also be in a better position to test this change with your branch of T-C*

ajs6f:
At CNI momentarily but will review and test against `trellis-cassandra`
this evening!

acoburn:
Great, thanks! If you're happy with this, I'd like to also review the methods in the `TrellisRequest` class, comparing them to how they are actually used in the various `.http.impl` classes. I suspect that we could simplify things a bit.

acoburn:
After re-reading the RESTeasy documentation linked to from above, I have updated this to proactively extract values from the context-derived objects before making use of them in a separate thread.

The only exception is the `Request` object, which is used to call `Request::evaluatePreconditions` (to check the `If-None-Match` and related headers). I am not sure what to do with that. The solution here is to just wrap this in a `null` check in the event that the processing thread no longer has access to the `request` object, in which case the cache-related checks will just be skipped over. That's not an ideal solution, though it's (arguably) better than throwing a NPE.

ajs6f:
Weirdness-- I'm getting test failures because there are too many slashes getting inserted. IOW I'm seeing things like 
```
DEBUG [pool-11-thread-1] 2018-12-12 10:16:43,325 CassandraResourceService.java:113 - Resource <trellis:data//> was not found
DEBUG [pool-11-thread-2] 2018-12-12 10:16:43,325 CassandraResourceService.java:113 - Resource <trellis:data///d627f265-64a2-4e40-8b2f-e5dad848b3c7> was not found
```
and to `trellis-c*`, that's a different URI (maybe it shouldn't be, but for now it is).
Any way this could be related? I'm wondering if we are accidentally using different URI resolution now because of the changes? Do I need to insert string munging to normalize paths (I'd really rather not, for performance)?

acoburn:
That's probably related, since the `Request::getPath` method gets changed around. I'm still working on the implementation for this, but I'll see if I can replicate that particular issue.

acoburn:
If you can test a `POST` to the root location again, that would be great. I recall, when using CXF, the `UriInfo::getPath` method returned a different value for the root resource depending on the framework being used (i.e. "" with Jersey and "/" with CXF), I'm guessing the same applies to RESTeasy. I did change the `uriInfo.getPath()` method call (from the earlier commit) to `uriInfo.getPathParameters().getFirst("path")`, which is more in line with how it is currently working in `master` (i.e. `@PathParam("path") String path`).

ajs6f:
Going to test now...

PR #329, opened 2018-12-12
Add hints to the BinaryService layer

### LINKED ISSUE #328: Binary hints
Well, we took out the ability to offer persistence hints to the backend, waiting for use cases, and I finally have one. In order to potentially use `trellis-cassandra` to distribute ML tasks to bitstreams (a major focus of DRAS-TIC) we need to be able to choose an appropriate chunk-length for bitstream storage. `t-c*` can vary that on a per-resource basis, so no problem there, but there is currently no way to get that info to the backend at resource creation/replacement...

We'd need some work in the Java for this, but even more, we need to think about a way this makes sense in HTTP. I don't like the idea of magic headers, but maybe we can offer a dedicated single header. But I would much rather use a better idea!

acoburn:
Resolves #328

This passes HTTP headers into the BinaryService, which can
be used as a form of storage hint. They can also be entirely ignored.

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/BinaryService.java
+++ b/core/api/src/main/java/org/trellisldp/api/BinaryService.java
@@ -36,10 +36,10 @@ public interface BinaryService extends RetrievalService<Binary> {
      *
      * @param metadata the binary metadata
      * @param stream the content
-     * @param headers HTTP request headers
+     * @param hints any hints for storing the binary data
      * @return the new completion stage
      */
-    CompletableFuture<Void> setContent(BinaryMetadata metadata, InputStream stream, Map<String, List<String>> headers);
+    CompletableFuture<Void> setContent(BinaryMetadata metadata, InputStream stream, Map<String, List<String>> hints);
 
     /**
      * Set the content for a binary object using a digest algorithm.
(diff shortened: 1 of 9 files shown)


core/api/src/main/java/org/trellisldp/api/BinaryService.java
```diff
@@ -34,9 +36,10 @@
      *
      * @param metadata the binary metadata
      * @param stream the content
+     * @param headers HTTP request headers
```
ajs6f: I would simply call this "hints". I can _easily_ imagine writing some Java code against the core and using this param for some totally non-HTTP metadata.

core/api/src/main/java/org/trellisldp/api/BinaryService.java
```diff
@@ -45,12 +48,13 @@
      * @param metadata the binary metadata
      * @param stream the context
      * @param algorithm the digest algorithm
+     * @param headers HTTP request headers
```
ajs6f: See above

acoburn:
The HTTP headers are already added [here](https://github.com/trellis-ldp/trellis/pull/329/files#diff-dbdc9e64a5b4bf885ce109b14a499b53R241) and [here](https://github.com/trellis-ldp/trellis/pull/329/files#diff-dbdc9e64a5b4bf885ce109b14a499b53R253).

Good idea about the javadocs.

ajs6f:
A last question: I'm getting `null` in this parameter now-- is that the case because there are no "no-Trellis" headers, and if so, wouldn't it be better to send an empty `Map`?

acoburn:
That's interesting. I'll take a look

ajs6f:
Thanks!

acoburn:
I bet that I will need to _copy_ the values [here](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/core/TrellisRequest.java#L74-L78) rather than just using a pointer. The issue (I am guessing) is that the underlying `HttpHeaders` object goes out of scope when it changes to a different thread, which means the reference to the corresponding header Map also goes out of scope.

ajs6f:
That sounds right-- it's exactly what happened in the asynchrony story.

acoburn:
I'll put that into a PR and you can give it a try

PR #330, opened 2018-12-13
Copy request headers into separate map

acoburn:
Follow-in to #329 

The important thing to note about this is that the keys of the resulting `MultivaluedMap` will all be **lower case**


ajs6f:
Hang on, @acoburn -- I need to investigate a problem in `t-c*`: this PR may not be as necessary as I led you to think.

PR #333, opened 2018-12-14
Less boxing in most cases

ajs6f:


acoburn:
This is good. I just added a commit to the file module, which should make the build pass.

ajs6f:
Does the build break on unnecessary boxing?

acoburn:
Yeah, the trellis-file module wouldn't compile, but it should now.

ajs6f:
Yes, let's.

PR #335, opened 2018-12-14
A lot less autoboxing/unboxing

acoburn:
Continuation of #333


ajs6f:
Are you leaving `BinaryMetadata` alone because we're using `null` to signal no-size in the ctor?

acoburn:
I didn't touch `BinaryMetadata` because `Optional<T>` requires an `Object` -- it won't take primitive types. And since that leaves us with `Optional<Long>`, I just kept the `Long` types all through that class. But there may be a better way around this issue, which is to remove the `::getSize` method entirely -- I don't think it's used anywhere, but we should discuss that separately.

Issue #349, opened 2019-01-31
WebDAV extension


ajs6f:
Providing WebDAV access to repository contents via a new JAX-RS bean would be practical and very attractive to many users. As discussed in an email thread, several new HTTP methods would be supported:

- COPY: A `ResourceService::get` followed by a `::replace` (or `::create`).
- MOVE: Same as COPY but with a :`:delete` at the end of the sequence.
- MKCOL: An `LDP:BasicContainer` via `ResourceService::create`.
- PROPFIND: `ResourceService.get().thenApply(someFiltering)`.
- PROPPATCH: Applied directly to the graph then persisted via `ResourceService::replace`.
- LOCK / UNLOCK: Optional, probably best not to implement.

acoburn:
Just an implementation note, as I start to think about this:

In order for a WebDAV client to properly navigate the LDP containment hierarchy as files and directories, we'll need a JAX-RS response filter for any LDP Container, which collects the containment triples and transforms these into the appropriate XML format for [collection resources](https://tools.ietf.org/html/rfc4918#section-5.2). For any LDP RDFSource/NonRDFSource, the regular GET payload can be returned without any modification.

We may also want to add a response filter for the `OPTIONS` method so that the additional `Allow` headers can be presented to clients (possibly for `GET`, too).

acoburn:
Another consideration: does the trellis-webdav module support the `Depth` header (for `GET`) and/or recursive `DELETE`. As a general principle, I'm not in favor of recursive delete, but in a WebDAV context, it makes a lot of sense (it's actually really easy to implement in a JAX-RS response filter that is isolated to this particular module).

ajs6f:
I have no problem with recursive `DELETE` in this context as long as there are no atomicity guarantees around it. I see none required by the RFC.

PR #353, opened 2019-02-09
Move HttpSession from impl to core package

### LINKED ISSUE #352: Move HttpSession from impl to core package
This will make it easier to re-use that class in other modules.

acoburn:
Resolves #352
Related to #349

This change will make it easier to re-use this internal Class in the
WebDAV module.


acoburn:
Superseded by 9cf08a3ab69b85f3895e901f6143cd5bdb8be800

PR #355, opened 2019-02-10
Add a no-op AccessControlService

### LINKED ISSUE #354: Add a no-op AccessControlService
This service would, effectively, grant access to all resources.

acoburn:
Resolves #354

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/NoopAccessControlService.java
+++ b/core/api/src/main/java/org/trellisldp/api/NoopAccessControlService.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


core/api/src/main/java/org/trellisldp/api/NoopAccessControlService.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.commons.rdf.api.RDF;
+
+/**
+ * A no-op AccessControlService implementation.
+ */
+public class NoopAccessControlService implements AccessControlService {
+
+    private static final RDF rdf = getInstance();
+    private static final String URI = "http://www.w3.org/ns/auth/acl#";
+    private static final Set<IRI> modes = asList("Control", "Read", "Write", "Append").stream()
```
ajs6f: I know it's Guava, but seems like ` ImmutableSet.of(rdf.createIRI("Control"), ...)` is a lot clearer here and it is also guaranteed immutable.

acoburn: I don't really want to have the API depend on Guava. What if we use something like `Collections.unmodifiableSet(...)`. The creation of it isn't as clean as with Guava, but it doesn't add the extra dependency.

ajs6f: Oh, I didn't realize it wasn't. For some reason I thought it came along with some Commons library. No, in that case `Collections.unmodifiableSet` would be great.

PR #360, opened 2019-02-12
Remove support for Digest/Want-Digest

### LINKED ISSUE #359: Consider removing support for Want-Digest
The `Digest` and `Want-Digest` features stem from an arguably problematic assumption that an HTTP-based software stack should be responsible for checking the bits in a persistence layer. There are better and more efficient ways to do this, such as checking the persistence layer directly.

acoburn:
Resolves #359


acoburn:
I've got another commit on its way -- some dangling references in test configuration files.

I also realize that by removing this, we also plug a potential DoS vector. For example, given a publicly available large binary, a client could repeatedly issue `HEAD` requests with `Want-Digest`. It would easily trigger a large CPU spike.

ajs6f:
:confounded:

PR #361, opened 2019-02-14
Make `WebACService` a CDI `@Alternative`

ajs6f:
Make `WebACService` a CDI `@Alternative` to avoid ambiguity with `NoopAccessControlService`


ajs6f:
@acoburn We already did this on a branch-- just getting it in here on `master`.

PR #368, opened 2019-03-12
Convert Tamaya to Microprofile-Config

### LINKED ISSUE #366: Switch from Tamaya API to MicroProfile Config
Tamaya, while excellent software, does not seem likely to produce its own long-lasting API. What is more, it already can implement MicroProfile Config (including injection). We should switch to code against µP Config's API, using Tamaya to implement it.

acoburn:
Resolves #366 

This uses SmallRye for the µP-config implementation, which is the impl that Thorntail currently uses. It seems to be actively developed and widely used. If we want to change that to tamaya-microprofile we can do that later.


acoburn:
ping @ajs6f there's no need to merge this right away if it will disrupt any testing efforts in the T-C* implementation. OTOH, it would be nice to get this change in place.

ajs6f:
I was able this, but not load it into Eclipse (as usual), so no review. I'm pretty much done trying to use Gradle with Eclipse, which really minimizes my options for reviewing PRs of any length.

acoburn:
Well, I do really like Gradle, but perhaps we should talk about changing over to maven. I can say that, when it comes to microprofile, the Gradle tooling is somewhere between non-existent and seriously incomplete.

ajs6f:
Hm. I don't want to make a big change on the basis of my convenience. Given the nature of this current PR (and the fact that it builds fine for me) I'm okay to move it through and then talk about build stuff separately. But it could have knock-on effects for `t-c*`, so let me check those out first.

acoburn:
Yes, let's discuss maven/gradle separately. I was able to do a local `./gradlew clean install` with this PR and then successfully build t-c* (after adding a `tamaya-microprofile` dependency to the pom.xml file in ./webapp).

ajs6f:
That's what I would have expected, but are you okay to let this hover while I check to see how hard it would be to shift `t-c*1` over to µP-config at the same time?

acoburn:
Fine with me. Take your time.

ajs6f:
Hm, µP-config doesn't offer multiple names for config keys (such as `t-c*` uses to provide both idiomatic JVM system prop and env var names). I'll check Tamaya's impl and see if that makes for an easier transition.

acoburn:
I have found a number of differences between µP-config and Tamaya, especially the fact that a non-existent key in Tamaya returns a `null` value while it throws a `NoSuchElementException` in µP-config (which is why in this PR, I use the `Config::getOptionalValue` a lot. It seems that µP-config has some changes in store for its next version, particularly the `Config::access` method, which makes it much easier to set default values.

ajs6f:
Ok, I can confirm that adding deps for Smallrye Config _and_ the Eclipse-sourced µP-config API gave me a build for `t-c*`. I say let's roll this PR in and I will keep using Tamaya for a bit with `t-c*` until @gregjan and I can plan a good and non-disruptive set of changes. Because none of the testing is using Trellis config directly, it should not be an issue for Greg that we change  Trellis config.

gregjan:
Yes, I am using java -D properties only at the moment. So I imagine that it will be easy enough for me to adjust to any renaming.

PR #370, opened 2019-03-12
Use microprofile metrics

### LINKED ISSUE #369: Replace dropwizard metrics with microprofile metrics in HTTP layer
The HTTP layer should make use of the µP metrics annotations.

acoburn:
Resolves #369

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,6 @@ ext {
     javaxJmsVersion = '2.0.1'
     javaxManagementVersion = '1.1.2'
     jaxrsVersion = '2.1.1'
-    metricsVersion = '4.0.5'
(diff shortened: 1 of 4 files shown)


acoburn:
Given that these are annotations, a deployable system can just ignore them if there is no metrics implementation. As a case in point, the Dropwizard-based deployable still works fine even though it doesn't support µP metrics (though these metrics annotations are based directly on Dropwizard)

ajs6f:
Ok, good. I had a lot of annoyances because Dropwizard wants a lot of dependencies and I had to align versions across it and other "big" dependencies. Hopefully µP metrics will be a bit lighter.

PR #375, opened 2019-03-25
Return quads from Resource::stream

### LINKED ISSUE #374: Return type from Resource::stream
There are three variations to the `Resource::stream` method:

```java
Stream<Quad> stream();

Stream<Triple> stream(Collection<IRI>);

Stream<Triple> stream(IRI);
```

What if all of these methods returned a `Stream<Quad>`? I think this would make the interface more consistent.

acoburn:
Resolves #374

### DIFF
--- a/components/file/src/main/java/org/trellisldp/file/FileResource.java
+++ b/components/file/src/main/java/org/trellisldp/file/FileResource.java
@@ -111,8 +111,8 @@ public class FileResource implements Resource {
 
     @Override
     public boolean hasAcl() {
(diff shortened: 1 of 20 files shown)


acoburn:
OK, just be aware that the `TrellisUtils::toQuad` method is removed with this commit, so [this line](https://github.com/trellis-ldp/trellis-cassandra/blob/master/impl/src/main/java/edu/si/trellis/CassandraResource.java#L127) will need modification, as will [these methods](https://github.com/trellis-ldp/trellis-cassandra/blob/master/impl/src/main/java/edu/si/trellis/CassandraResource.java#L133-L151).

PR #378, opened 2019-03-28
Add configuration key for PUT containment

acoburn:
Resolves trellis-ldp/trellis-cassandra#57

With this change, resource creation via PUT can be configured to be either
contained or uncontained. By default, the configuration is uncontained.

The integration tests have been updated to consider this case.

The webdav component requires that creation via PUT is configured as
'contained' rather than 'uncontained', and there are some efficiency
gains for configuring it as 'contained', but there is a fallback in case
the configuration is different.

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/impl/PutHandler.java
+++ b/core/http/src/main/java/org/trellisldp/http/impl/PutHandler.java
@@ -143,9 +143,12 @@ public class PutHandler extends MutatingLdpHandler {
             throw new NotAcceptableException();
         }
 
-        if (createUncontained) {
-            ofNullable(resource).flatMap(Resource::getContainer).ifPresent(p -> setParent(parent));
-        } else {
+        // For operations that modify resources, the parent resource may need to be updated via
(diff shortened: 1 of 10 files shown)


core/http/src/main/java/org/trellisldp/http/impl/PutHandler.java
```diff
@@ -139,7 +143,11 @@ public ResponseBuilder initialize(final Resource parent, final Resource resource
             throw new NotAcceptableException();
         }
 
-        ofNullable(resource).flatMap(Resource::getContainer).ifPresent(p -> setParent(parent));
+        if (createUncontained) {
+            ofNullable(resource).flatMap(Resource::getContainer).ifPresent(p -> setParent(parent));
```
ajs6f: Can you leave an inline comment here for this? It's not obvious to me quite what is happening…

acoburn: Definitely. There are a few more things to clean up from this PR, which I'll add in another commit

PR #379, opened 2019-03-29
Improve handling of client-provided Slug header

### LINKED ISSUE #376: POST with invalid Slug header can lead to inaccessible resources
A POST request with a `Slug` header containing problematic characters can lead to inaccessible resources. For instance, creating a resource with the header `Slug: resource#foo` will successfully create a resource. The response headers will contain `Location: http://example.com/resource#foo`, but subsequent requests for `/resource` or `/resource#foo` result in `404 Not Found`. This is because the internal identifier will contain a hash and it will not be possible to locate that value

The same issue applies for Slug values with `?` as well as any percent-encoded values.

The solution is to either reject those requests and/or to strip off any trailing `#` or `?` sequences.

acoburn:
Resolves #376 

Slug values should be URL-decoded, so this adds a new `http.core.Slug` class for handling this step. The rest of the changes are, effectively, adding test cases.

There are also some behavior changes, namely:

1. if the `Slug` headers contains any `/` characters, those will be replaced with an underscore rather than causing the entire request to fail (which is the current behavior)
2. `Slug` headers with whitespace characters will have that whitespace also replaced with underscores.
3. `Slug` headers with a fragment IRI (`#foo`) will have that fragment stripped off
4. `Slug` headers with query parameters (`?bar=baz`) will have that stripped off

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/core/Slug.java
+++ b/core/http/src/main/java/org/trellisldp/http/core/Slug.java
@@ -38,6 +38,7 @@ import org.slf4j.Logger;
 public class Slug {
 
     private static final Logger LOGGER = getLogger(Slug.class);
+    private static final URLCodec DECODER = new URLCodec();
 
     private final String slugValue;
 
@@ -46,7 +47,7 @@ public class Slug {
      * @param value the value of the Slug header.
      */
     public Slug(final String value) {
-        this.slugValue = replaceChars(requireNonNull(value, "Must be a non-null value!"));
+        this.slugValue = cleanSlugString(requireNonNull(value, "Must be a non-null value!"));
     }
 
     /**
@@ -68,15 +69,16 @@ public class Slug {
 
     private static Optional<String> decodeSlug(final String value) {
         try {
-            final URLCodec decoder = new URLCodec();
-            return of(decoder.decode(value));
+            return of(DECODER.decode(value));
         } catch (final DecoderException ex) {
             LOGGER.warn("Error decoding slug value, ignoring header: {}", ex.getMessage());
         }
         return empty();
     }
 
-    private static String replaceChars(final String value) {
+    private static String cleanSlugString(final String value) {
+        // Remove any fragment URIs and query parameters
+        // Then trim the string and replace any remaining whitespace or slash characters with underscores
         return value.split("#")[0].split("\\?")[0].trim().replaceAll("[\\s/]+", "_");
     }
 }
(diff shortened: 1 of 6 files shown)


core/http/src/main/java/org/trellisldp/http/core/Slug.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param value the raw value of the HTTP header, may be null
+     * @return a Slug object with a decoded value or null
+     */
+    public static Slug valueOf(final String value) {
+        return ofNullable(value).flatMap(Slug::decodeSlug).map(Slug::new).orElse(null);
+    }
+
+    private static Optional<String> decodeSlug(final String value) {
+        try {
+            final URLCodec decoder = new URLCodec();
```
ajs6f: `URLCodec` is threadsafe, so this could be a field.

core/http/src/main/java/org/trellisldp/http/core/Slug.java
```diff
@@ -0,0 +1,82 @@
...
+            return of(decoder.decode(value));
+        } catch (final DecoderException ex) {
+            LOGGER.warn("Error decoding slug value, ignoring header: {}", ex.getMessage());
+        }
+        return empty();
+    }
+
+    private static String replaceChars(final String value) {
```
ajs6f: This doesn't just replace chars, it slices and dices, so maybe a more expressive name?

core/http/src/main/java/org/trellisldp/http/core/Slug.java
acoburn: Good point. This is now a static field.

core/http/src/main/java/org/trellisldp/http/core/Slug.java
acoburn: I'm calling it `cleanSlugString` now and I've included some in-line documentation explaining what it is doing.

core/http/src/main/java/org/trellisldp/http/core/Slug.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Get a Slug object from a decoded string value.
+     * @param value the raw value of the HTTP header, may be null
+     * @return a Slug object with a decoded value or null
+     */
+    public static Slug valueOf(final String value) {
+        return ofNullable(value).flatMap(Slug::decodeSlug).map(Slug::new).orElse(null);
+    }
+
+    private static Optional<String> decodeSlug(final String value) {
```
ajs6f: If this returns a plain `String`, isn't the use above simpler:
```
         return ofNullable(value).map(Slug::decodeSlug).map(Slug::new).orElse(null);
```

acoburn: Good point -- I just added another commit to simplify this

PR #384, opened 2019-04-03
Add a baseUrl parameter to the WebAcFilter class

### LINKED ISSUE #382: WebAC filter does not respect baseURL override
As reported on the [Trellis email list](https://groups.google.com/d/msg/trellis-ldp/026H9vmTjs4/oA9aJ1q0AwAJ) by @mjgiarlo, when a custom baseURL is in use, the WebAC filter does not take this into consideration when generating a link header in the response. For example, with a custom base url of `http://example.com/`, the response link header might contain, instead:

```http
Link: <http://localhost/resource?ext=acl>; rel="acl"
```

acoburn:
Resolves #382

@ajs6f note that the signature for WebAcFilter (and webdav authZ filter) changes and may require adjustment to the downstream T-C* implementation.

### DIFF
--- a/components/app/src/main/java/org/trellisldp/app/AbstractTrellisApplication.java
+++ b/components/app/src/main/java/org/trellisldp/app/AbstractTrellisApplication.java
@@ -142,7 +142,8 @@ public abstract class AbstractTrellisApplication<T extends TrellisConfiguration>
                 .ifPresent(challenges::add);
             of(config.getAuth().getBasic()).filter(BasicAuthConfiguration::getEnabled).map(x -> "Basic")
                 .ifPresent(challenges::add);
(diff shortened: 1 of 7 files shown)


acoburn:
Note: this should be backported to the trellis-0.8.x branch with a minor change such that the constructor signatures continue to exist in their present form (an additional constructor with the baseURL would be added)

ajs6f:
Hm, @acoburn, I don't implement those types in any sense-- do you mean that the ctors change and therefore injection might change?

acoburn:
@ajs6f yeah, just that the ctors change, and so in the event that you were calling those directly this might require a change. However, because you're using CDI injection, it's quite possible that you won't need to make any changes.

PR #385, opened 2019-04-03
Improve URI creation from baseUrl

### LINKED ISSUE #383: Custom baseURL without trailing slash can generate malformed timemap URLs
As reported on the [Trellis email list](https://groups.google.com/d/msg/trellis-ldp/026H9vmTjs4/oA9aJ1q0AwAJ) by @mjgiarlo, the timemap response has some malformed URIs, but this only seems to appear when a custom baseURL is in use and that custom URL does not contain a trailing slash.

While a work-around would be to simply add a trailing slash to the baseURL configuration, it seems overly picky to require a trailing slash on a baseURL. The code used to generate these URLs should allow for this possibility and adjust the output accordingly.

acoburn:
Resolves #383

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/impl/MementoResource.java
+++ b/core/http/src/main/java/org/trellisldp/http/impl/MementoResource.java
@@ -146,7 +146,7 @@ public final class MementoResource {
      */
     public ResponseBuilder getTimeGateBuilder(final SortedSet<Instant> mementos, final TrellisRequest req,
             final String baseUrl) {
-        final String identifier = getBaseUrl(baseUrl, req) + req.getPath();
+        final String identifier = fromUri(getBaseUrl(baseUrl, req)).path(req.getPath()).build().toString();
         return status(FOUND)
             .location(fromUri(identifier + "?version=" + req.getDatetime().getInstant().getEpochSecond()).build())
             .link(identifier, ORIGINAL + " " + TIMEGATE)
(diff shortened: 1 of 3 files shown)


core/http/src/main/java/org/trellisldp/http/impl/MementoResource.java
```diff
@@ -107,7 +107,7 @@ public ResponseBuilder getTimeMapBuilder(final SortedSet<Instant> mementos, fina
             final String baseUrl) {
 
         final List<MediaType> acceptableTypes = req.getAcceptableMediaTypes();
-        final String identifier = getBaseUrl(baseUrl, req) + req.getPath();
+        final String identifier = fromUri(getBaseUrl(baseUrl, req)).path(req.getPath()).build().toString();
```
mjgiarlo: I haven't spent much time at all in the Trellis code, but I wonder whether making this change without a related change in the test suite (to prevent regressions) might come back to bite us. 

Thanks for addressing these issues so quickly, @acoburn!

acoburn: Yes, adding some regression tests for this is a very good idea :-)

acoburn: Additional tests are on the way -- with the new tests, the timemap tests failed (expectedly) when run against the current `master` branch, but they succeed with this modification in place.

PR #387, opened 2019-04-04
Adjust PATCH requests to operate on externalized IRIs

### LINKED ISSUE #386: PATCH requests with absolute in-domain URIs are not correctly mapped to resource graphs
Given a resource:

```turtle
<> a ex:Foo ;
    dc:title "A title" .
```

The SPARQL-Update request to delete these triples only works with relative URIs:

```
DELETE DATA {
  <> dc:title "A title" .
}
```

With absolute URIs, the delete statement does not match the resource triples (and the triples are not removed). This is due to the fact that when the resource triples are loaded into a `Graph` in preparation for running the `IOService::update` operation, the triples are not first converted from their internal representation to an external representation and the context of the update query uses the internal IRI value (rather than the external resource URL).

acoburn:
Resolves #386

This also adds an integration test to check this behavior

### DIFF
--- a/components/test/src/main/java/org/trellisldp/test/LdpRdfTests.java
+++ b/components/test/src/main/java/org/trellisldp/test/LdpRdfTests.java
@@ -265,7 +265,8 @@ public interface LdpRdfTests extends CommonTests {
             assertAll("Check an LDP-RS resource", checkRdfResponse(res, LDP.RDFSource, null));
         }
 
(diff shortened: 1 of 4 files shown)


acoburn:
As a note: this should be back-ported to the `trellis-0.8.x` branch.

mjgiarlo:
Thanks again, @acoburn!

PR #390, opened 2019-04-18
Various small syntax cleanups

ajs6f:
Mostly removing the use of Objects::nonNull and Objects::isNull, with a few other bits thrown in.

This PR only covers the stack up through `-http` and ignores all the backends and optional components. They can be handled in separate PRs to avoid crazy levels of line-change too fast.

### DIFF
--- a/components/agent/src/main/java/org/trellisldp/agent/SimpleAgentService.java
+++ b/components/agent/src/main/java/org/trellisldp/agent/SimpleAgentService.java
@@ -13,7 +13,6 @@
  */
 package org.trellisldp.agent;
 
-import static java.util.Objects.nonNull;
 import static org.trellisldp.api.TrellisUtils.getInstance;
 
 import org.apache.commons.rdf.api.IRI;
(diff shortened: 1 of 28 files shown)


ajs6f:
The Travis build failed for nothing to do with this PR:
```
The command "~/bin/install-jdk.sh --target "/home/travis/openjdk11" --workspace "/home/travis/.cache/install-jdk" --feature "11" --license "GPL" --cacerts" failed and exited with 1 during .
```

acoburn:
The Java11 downloader/installer tends to fail for a while right around the time of an update (11.0.3 was just released). It should be operational again soon, but that's pretty much out of our hands.

ajs6f:
I figured it was some silliness like that.

acoburn:
LGTM. I'll double-check the JDK11 build tomorrow. By then it should have sorted itself out

ajs6f:
Sounds good. I'll be out-of-touch tomorrow but back on Sunday. Feel free to reject anything you think makes the code _less_ idiomatic, because my opinions have been known to be a bit surreal at times.

PR #392, opened 2019-04-23
[WIP] Implement websocket support

### LINKED ISSUE #29: Add a WebSocket-based notification system
The SOLID specification recommends using WebSockets for notifications, and I think this would be a good way to provide an embedded message broker for Trellis.

There are a few outstanding questions to resolve w/r/t implementation. First, SOLID suggests making websocket endpoints available on each resource. This seems like a very interesting idea, as the notifications can be scoped to a particular resource or container of resources.

  1. If a websocket is initiated at resource `/foo`, should notifications from _all_ contained resources be made available (i.e. recursively?) or just those that that relate to that resource/container?
  2. If `1.` is answered in the negative, is there a resource location where _all_ notifications can be accessed?

acoburn:
Resolves #29

This adds basic websocket support as defined by [Solid](https://github.com/solid/solid-spec/blob/master/api-websockets.md). Before merging, this should also support Authorization and should be more thoroughly tested.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,6 +47,7 @@ ext {
     kafkaVersion = '2.2.0'
     mustacheVersion = '0.9.6'
     rabbitMqVersion = '5.7.0'
+    websocketApiVersion = '1.1'
     validationVersion = '2.0.1.Final'
 
     /* Testing */
--- a/components/websocket/build.gradle
+++ b/components/websocket/build.gradle
@@ -1,0 +1,49 @@
+apply plugin: 'java-library'
+apply plugin: 'biz.aQute.bnd.builder'
+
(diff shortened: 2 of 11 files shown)


ajs6f:
@acoburn Do you want to just hold this open until you feel like it is robust enough to merge, or are we going to wait on use cases (or did y'all come up with use cases after I left yesterday's call)?

acoburn:
I'd like to hold this open until the implementation is more robust. I mostly wanted to get this off of a private branch on my local machine.

ajs6f:
👍

acoburn:
Closing per this [discussion](https://github.com/trellis-ldp/trellis/issues/29#issuecomment-508160832)

PR #399, opened 2019-05-07
Use LDP type of the deleted resource in events

### LINKED ISSUE #398: What rdf:type should DELETE notifications use
@mjgiarlo brought this up on the mailing list. When a Trellis resource is deleted, the notification that is emitted has the `ldp:Resource` type. It seems that using the type of the deleted resource would make more sense. That is, if a basic container is deleted, then the type would be `ldp:BasicContainer`.

acoburn:
Resolves #398

When a DELETE request completes, the notification will contain the
rdf:type of the deleted resource, rather than the generic ldp:Resource.

### DIFF
--- a/components/test/src/main/java/org/trellisldp/test/EventTests.java
+++ b/components/test/src/main/java/org/trellisldp/test/EventTests.java
@@ -215,7 +215,7 @@ public interface EventTests extends CommonTests {
         try (final Response res = target(resource).request().header(AUTHORIZATION, buildJwt(agent2, getJwtSecret()))
                 .delete()) {
             assertEquals(SUCCESSFUL, res.getStatusInfo().getFamily(), "Verify a successful LDP-RS DELETE response");
-            assertAll("Check the LDP-BC parent", checkResourceParentLdpBC(resource, agent2, AS.Delete, LDP.Resource));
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@acoburn Once merged, do you think this could be backported for a 0.8.2 release? (Depends, I guess, on the extent to which you consider this a bugfix in the semver sense...)

mjgiarlo:
@acoburn @ajs6f and thanks, btw! I am continually impressed by your responsiveness.

acoburn:
I don’t see any problem back-porting this to a 0.8.2. None of the Java APIs
change with this, so even from the perspective of SemVer, it’s not an issue.

mjgiarlo:
@acoburn Excellent. Very glad to hear it!

mjgiarlo:
@acoburn That would be fabulous. Thank you!

mjgiarlo:
@acoburn I see the release of 0.8.2 has begun. Can you ping me when I can pull this in an updated `trellis-ext-db` container?

Thanks much!

acoburn:
@mjgiarlo The `trellis-ext-db:0.2.2` container is now available. Enjoy :-)

https://hub.docker.com/r/trellisldp/trellis-ext-db/tags

mjgiarlo:
@acoburn excellent! Thank you, sir.

PR #405, opened 2019-05-22
Adding no-arg ctors to satisfy CDI

ajs6f:
https://github.com/trellis-ldp/trellis-cassandra/pull/65
https://github.com/gregjan/trellis-cassandra/pull/1

### DIFF
--- a/components/webdav/src/main/java/org/trellisldp/webdav/TrellisWebDAV.java
+++ b/components/webdav/src/main/java/org/trellisldp/webdav/TrellisWebDAV.java
@@ -142,6 +142,13 @@ public class TrellisWebDAV {
         this(services, getConfig());
     }
 
+    /**
+     * Used only to support CDI.
+     */
+    TrellisWebDAV() {
+        this(null);
+    }
+
     private TrellisWebDAV(final ServiceBundler services, final Config config) {
         this(services, config.getOptionalValue(CONFIG_HTTP_BASE_URL, String.class).orElse(null));
     }
--- a/components/webdav/src/main/java/org/trellisldp/webdav/TrellisWebDAVRequestFilter.java
+++ b/components/webdav/src/main/java/org/trellisldp/webdav/TrellisWebDAVRequestFilter.java
@@ -73,7 +73,7 @@ public class TrellisWebDAVRequestFilter implements ContainerRequestFilter {
     public TrellisWebDAVRequestFilter(final ServiceBundler services) {
         this(services, getConfig());
     }
-    
+
     /**
      * Use only to support CDI.
      */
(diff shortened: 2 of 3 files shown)


components/webdav/src/main/java/org/trellisldp/webdav/TrellisWebDAV.java
```diff
@@ -142,6 +142,13 @@ public TrellisWebDAV(final ServiceBundler services) {
         this(services, getConfig());
     }
 
+    /**
+     * Used only to support CDI.
+     */
+    TrellisWebDAV() {
+        this(null);
```
acoburn: In this case, is the `ServiceBundler` injected through some other means? Or is this merely to have a no-arg ctor available?

ajs6f: The latter. It is frankly ugly, but it's CDI. There are impl-specific workarounds, but I don't want to do that, do you?

Briefly, because of the scoping of the use of this guy (and the others) in JAX-RS, CDI wants to use a proxy for the bean. It can use the `@Inject`-annotated fielded ctor for the impl of that proxy, but in order to build the proxy itself, it needs the no-arg ctor.

acoburn: Got it. Then this sounds good. I think there's a pesky checkstyle issue on line 76 of the TrellisWebDAVRequestFilter class, but otherwise LGTM.

ajs6f: I'll check it now.

components/webdav/src/main/java/org/trellisldp/webdav/TrellisWebDAVRequestFilter.java
```diff
@@ -73,6 +73,13 @@
     public TrellisWebDAVRequestFilter(final ServiceBundler services) {
         this(services, getConfig());
     }
+
```
acoburn: There's some checkstyle nonsense going on here. Probably some trailing whitespace.

ajs6f:
Just got that whitespace.

gregjan:
Okay, this implies to me that the TrellisHttpResource.initialize() method will be called @PostConstruct regardless of the other services in the mix b/c injection is happening. Correct?

ajs6f:
All `@PostConstruct` are guaranteed to be called by the container before the application is in runtime, so yes, it will.

ajs6f:
Just FYI, from the `@PostConstruct` Javadocs:
```
* ... This
 * method MUST be invoked before the class is put into service. This
 * annotation MUST be supported on all classes that support dependency
 * injection. The method annotated with PostConstruct MUST be invoked even
 * if the class does not request any resources to be injected.
```
I find them extremely reliable.

PR #413, opened 2019-06-09
sets priorities for alternative CDI beans

gregjan:
This PR can be more narrowly focused based on feedback, especially if it is adding too much logging. The gist of it is that I needed to set Priority and Alternative annotations in order to get CDI to load the correct implementation, instead of Noops, when present.

### DIFF
--- a/auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
+++ b/auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
@@ -45,7 +45,7 @@ import org.slf4j.Logger;
 public class BasicAuthFilter implements ContainerRequestFilter {
 
     private static final Logger LOGGER = getLogger(BasicAuthFilter.class);
-    
+
     /** The configuration key controlling the location of the basic auth credentials file. **/
     public static final String CONFIG_AUTH_BASIC_CREDENTIALS = "trellis.auth.basic.credentials";
 
--- a/auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
+++ b/auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
@@ -87,7 +87,7 @@ public class BasicAuthFilter implements ContainerRequestFilter {
     public BasicAuthFilter(final File file, final String realm) {
         this.file = file;
         this.challenge = "Basic realm=\"" + realm + "\"";
-        LOGGER.info("Configured BASIC Authentication Filter.");
+        LOGGER.info("Configured BasicAuthFilter.");
     }
 
     @Override
--- a/auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
+++ b/auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
@@ -18,6 +18,7 @@ import static javax.ws.rs.Priorities.AUTHENTICATION;
 import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
 import static javax.ws.rs.core.SecurityContext.BASIC_AUTH;
 import static org.eclipse.microprofile.config.ConfigProvider.getConfig;
+import static org.slf4j.LoggerFactory.getLogger;
 
 import java.io.File;
 import java.io.IOException;
@@ -32,13 +33,19 @@ import javax.ws.rs.NotAuthorizedException;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
 import javax.ws.rs.core.SecurityContext;
+import javax.ws.rs.ext.Provider;
+
+import org.slf4j.Logger;
 
 /**
  * A basic authentication filter using an Authorization HTTP header.
  */
+@Provider
 @Priority(AUTHENTICATION)
 public class BasicAuthFilter implements ContainerRequestFilter {
 
+    private static final Logger LOGGER = getLogger(BasicAuthFilter.class);
+    
     /** The configuration key controlling the location of the basic auth credentials file. **/
     public static final String CONFIG_AUTH_BASIC_CREDENTIALS = "trellis.auth.basic.credentials";
 
@@ -80,6 +87,7 @@ public class BasicAuthFilter implements ContainerRequestFilter {
     public BasicAuthFilter(final File file, final String realm) {
         this.file = file;
         this.challenge = "Basic realm=\"" + realm + "\"";
+        LOGGER.info("Configured BASIC Authentication Filter.");
     }
 
     @Override
--- a/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
@@ -44,8 +44,10 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
+import javax.annotation.Priority;
 import javax.enterprise.inject.Alternative;
 import javax.inject.Inject;
+import javax.interceptor.Interceptor;
 
 import org.apache.commons.rdf.api.Graph;
 import org.apache.commons.rdf.api.IRI;
@@ -73,6 +75,7 @@ import org.trellisldp.vocabulary.VCARD;
  * @author acoburn
  */
 @Alternative
+@Priority(Interceptor.Priority.APPLICATION+10)
 public class WebACService implements AccessControlService {
 
     /** The configuration key controlling whether to check member resources at the AuthZ enforcement point. **/
@@ -135,6 +138,7 @@ public class WebACService implements AccessControlService {
         this.resourceService = requireNonNull(resourceService, "A non-null ResourceService must be provided!");
         this.cache = requireNonNull(cache, "A non-null Cache must be provided!");
         this.checkMembershipResources = checkMembershipResources;
+        LOGGER.info("Loading WebACService");
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
```diff
@@ -32,13 +33,19 @@
...
 
 /**
  * A basic authentication filter using an Authorization HTTP header.
  */
+@Provider
 @Priority(AUTHENTICATION)
 public class BasicAuthFilter implements ContainerRequestFilter {
 
+    private static final Logger LOGGER = getLogger(BasicAuthFilter.class);
+
```
acoburn: There's some trailing whitespace here, which the checkstyle config if complaining about.

auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
```diff
@@ -80,6 +87,7 @@ public BasicAuthFilter(final File file) {
     public BasicAuthFilter(final File file, final String realm) {
         this.file = file;
         this.challenge = "Basic realm=\"" + realm + "\"";
+        LOGGER.info("Configured BASIC Authentication Filter.");
```
acoburn: Adding logging here is fine, though I'd recommend setting it at `DEBUG` level. Ditto for the other logging methods.

notifications/jms/src/main/java/org/trellisldp/jms/JmsPublisher.java
```diff
@@ -41,6 +44,8 @@
  *
  * @author acoburn
  */
+@Alternative
+@Priority(Interceptor.Priority.APPLICATION+10)
```
acoburn: Checkstyle issue (add whitespace around the `+` operator)

notifications/jms/src/main/java/org/trellisldp/jms/JmsPublisher.java
```diff
@@ -107,6 +112,7 @@ public JmsPublisher(final Session session, final String queueName) throws JMSExc
      * @throws JMSException when there is a connection error
      */
     public JmsPublisher(final Session session, final String queueName, final boolean useQueue) throws JMSException {
+    	LOGGER.info("Loading JmsPublisher");
```
acoburn: Checkstyle issue: don't use a tab here.

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
```diff
@@ -73,6 +75,7 @@
  * @author acoburn
  */
 @Alternative
+@Priority(Interceptor.Priority.APPLICATION+10)
```
acoburn: Checkstyle issue: add whitespace around the `+` operator

ajs6f:
In this kind of work there is no such thing as "too much logging", there is just logging at the wrong detail level.

auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
ajs6f: I disagree-- this is initialization code, performed only once. I think it would nice to see a single positive message for each component that successfully initializes.

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
ajs6f: If there is any magic to 10 or not, it would be nice to have it broken out into a constant and given a comment to the correct effect.

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
```diff
@@ -135,6 +138,7 @@ public WebACService(final ResourceService resourceService,
         this.resourceService = requireNonNull(resourceService, "A non-null ResourceService must be provided!");
         this.cache = requireNonNull(cache, "A non-null Cache must be provided!");
         this.checkMembershipResources = checkMembershipResources;
+        LOGGER.info("Loading WebACService");
```
ajs6f: `Loading` could =>`Loaded` to match usage above.

core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
```diff
@@ -186,11 +186,11 @@ public void initialize() {
 
     private CompletionStage<Void> initialize(final IRI id, final Resource res, final Dataset dataset) {
         if (MISSING_RESOURCE.equals(res) || DELETED_RESOURCE.equals(res)) {
-            LOGGER.info("Initializing root container: {}", id);
+            LOGGER.info("Initializing root container and ACL: {}", id);
```
ajs6f: I'm not sure that we have to hardwire WebAC in this way.

gregjan:
And now I've learned how to run checkstyle locally and that is a good thing.

auth/basic/src/main/java/org/trellisldp/auth/basic/BasicAuthFilter.java
acoburn: Yes, @ajs6f makes a good point here. Let's leave it this at `INFO` level.

acoburn:
To the point about the No-op pattern. I can see two decent possibilities (there may well be others):

1. Leave the No-op implementations in the `trellis-api` but remove the `beans.xml` file from that jar so that autowiring does not happen. Wiring of a no-op class would have to be done explicitly at an outer layer of the application.

2. Move the no-op implementations into separate jars. In this way, autowiring would happen only if the jar is explicitly added to the classpath.

Do those seem like viable options?

ajs6f:
In the case of 1., how do the non-no-op impls get wired? Via `beans.xml`s in their jars?
In the case of 2., what if you want one no-op but not the others?

acoburn:
In the case of 1: some other `beans.xml` configuration would need to wire in that no-op implementation (I'm assuming here that such a thing is possible)

In the case of 2: each no-op implementation would need to be in its own jar (i.e. 1 jar == 1 class) -- so there would be a lot more jar files at play

ajs6f:
I'm wondering with increasing curiosity whether we might want to go straight to bean autodiscovery.

gregjan:
Just reading back into some CDI materials and it got me wondering if the Interceptor pattern might be suitable for the various filters, which would not necessarily warrant an abstract @Inject and Noop pattern. So you'd add a jar that contains the BasicAuthFilter and it would have an InterceptorBinding to the Application class and a @PostConstruct method..

ajs6f:
@gregjan CDI interceptors are a general mechanism for intercepting calls to beans themselves, for example for logging or audit info (not HTTP traffic). Are you thinking of JAX-RS interceptors, because they are used around message writing and reading. I really think a filter is the right kind of component here, but I agree that we don't need a no op pattern here. In fact, we should be unwinding that pattern, not adding more of it.

gregjan:
@ajs6f I'm thinking more of the kind of interceptor described in [2.7](https://github.com/javaee/interceptors-spec/blob/master/1.2/Intercept.pdf). A bean might do something on @PostConstruct of the TrellisApplication, at which point the TA would already have been injected and finished it's own @PostConstruct method. So the "extension" bean would then be able to consume it's own injected dependencies (a ResourceService, for example) and perform it's initialization steps (add an ACL at the root).

gregjan:
It looks to me like interceptor bindings can be inherited, so we could define an extension abstract class with the right bindings and some callback or "register me" methods. I'm just spitballing here. It's been awhile since I've thought about wiring. This would likely require an abstract application class of some kind and a bit of contract around core/extension.

gregjan:
@ajs6f Is this what you mean by [bean autodiscovery](https://elmland.blog/2017/09/11/boot-cdi-2-javase/)?

ajs6f:
How would a bean do something triggered by the `@PostConstruct` method on another bean? Are you thinking of using [CDI Events](https://docs.oracle.com/javaee/7/tutorial/cdi-adv005.htm#GKHIC)?

gregjan:
I was thinking about [Intercepting Lifecycle Callback Events](https://javaee.github.io/tutorial/interceptors002.html), where the Interceptor is a separate class from the Interceptor target. So the extension class would be an interceptor that targets the TrellisApplication's PostConstruct lifecycle event.

ajs6f:
@gregjan That's autodiscovery for CDI, which we are using now (notice that I don't declare any beans anwhere, I just let the container discover what's needed). I'm talking about adding JAX-RS bean autodiscovery, aka annotation scanning.

ajs6f:
Wait, so the WebAC bean (and any other optional module) would intercept the `@PostConstruct` method on `TrellisHttpResource` in order to run its own initialization code?

gregjan:
Aha, so it would find all the relevant container filters in the classpath.. Those things that we don't want to have the application necessarily care about or have a property.

ajs6f:
Part of what we would have to do is make sure that JAX-RS and CDI inject the right things into the right places, and that's exactly where µProfile comes in. So a filter should be able to use CDI-managed beans (e.g. `ResourceService`).

acoburn:
The changes I suggested have all been addressed. As for the larger design issue, that goes a little beyond the scope of this change: should we move forward with merging this now and create a separate ticket for sorting out the issues @ajs6f brought up, or shall we try to figure out a solution as part of this PR?

ajs6f:
I'm not going to lay down on the tracks for this one, but I really am not very comfortable accepting https://github.com/trellis-ldp/trellis-cassandra/pull/66 (which hardwires WebAC into `trellis-cassandra`), so if this PR has intrinsic merit beyond supporting https://github.com/trellis-ldp/trellis-cassandra/pull/66, then let's merge it straightaway. Otherwise, I'd like to take a little more time and think about the larger wiring and optionality question. I guess I'm tired of kicking that can down the road. @gregjan Are you blocked now? Can we take that time now or should we put it off until you get more testing done?

ajs6f:
@gregjan, if the no-op impls were all labelled as `@Alternative` and the "substantive" impls were all labelled `@Default`, would you be having this problem? Wouldn't it be that you would just bring in WebAC and JMS in Maven and then CDI would do the rest? Would that let us kick the can a little further?

acoburn:
Superseded by #416

PR #415, opened 2019-06-13
Restrict audit stream to users with acl:Control

### LINKED ISSUE #414: Control access to audit stream
The audit stream on a given resource is currently accessible to any user with `acl:Read` access to a resource. That access should be restricted to users with `acl:Control` privileges. This can easily be implemented by adjusting the request filter in the `org.trellisldp.http.WebAcFilter` class.

acoburn:
Resolves #414

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
+++ b/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
@@ -26,6 +26,7 @@ import static org.slf4j.LoggerFactory.getLogger;
 import static org.trellisldp.api.TrellisUtils.TRELLIS_DATA_PREFIX;
 import static org.trellisldp.api.TrellisUtils.getInstance;
 import static org.trellisldp.http.core.HttpConstants.CONFIG_HTTP_BASE_URL;
+import static org.trellisldp.http.core.HttpConstants.PREFER;
 import static org.trellisldp.http.core.HttpConstants.SESSION_PROPERTY;
 
 import java.io.IOException;
@@ -52,6 +53,7 @@ import org.trellisldp.api.AccessControlService;
 import org.trellisldp.api.Session;
 import org.trellisldp.http.core.HttpConstants;
 import org.trellisldp.http.core.HttpSession;
+import org.trellisldp.http.core.Prefer;
 import org.trellisldp.vocabulary.ACL;
 import org.trellisldp.vocabulary.Trellis;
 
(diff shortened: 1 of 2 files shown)


core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
```diff
@@ -149,6 +151,14 @@ public void filter(final ContainerRequestContext req, final ContainerResponseCon
         }
     }
 
+    private boolean reqAudit(final ContainerRequestContext ctx) {
+        final Prefer prefer = Prefer.valueOf(ctx.getHeaderString(PREFER));
```
ajs6f: Just a thought, maybe change to `Optional<Prefer> Prefer::valueOf`? Or add a method under a new name, since `static valueOf` is pretty well-known from `enum`s.

acoburn: `Prefer::valueOf` will need to stay as it is in order to play nicely with the JAX-RS header parsing mechanism. Plus, aren't we trying to avoid all those `Optional` types? :-)

ajs6f: Not at all! This is [the exact use case](https://stackoverflow.com/a/26328555) for which they were meant. In any event, it was just a thought-- no need to hold this PR up.

acoburn: I agree with that use case. What I'm referring to, though, is the use of [`javax.ws.rs.HeaderParam`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/HeaderParam.html). Note the third item in the list of requirements:

> Have a static method named `valueOf` or `fromString` that accepts a single `String` argument (see, for example, `Integer.valueOf(String)`)

ajs6f: Sure, that's why I would "add a method under a new name" a la `Optional<Prefer> parse(String)` or the like. But it's that much more code, so unless you expect to do this a lot, probably easier to just do what you do here.

acoburn: Right, I don't expect to be doing this sort of thing very much.

acoburn:
@ajs6f do you have any additional thoughts on this PR? Or shall I go ahead an merge this?

ajs6f:
No, not at all-- I was fine to merge it from the beginning.

Issue #400, opened 2019-05-09
Deprioritize and "alternative-ize" no-op component implementations


ajs6f:
For best CDI ergonomics, the many no-op implementations of components (e.g. `NoopAccessControlService` or `NoopAuditService`) should be annotated as CDI `@Alternative`s and given a low priority for wiring. This will make it much easier to add substantive implementations that can auto-wire and avoid requiring deployers to ever tinker with `beans.xml` files.

ajs6f:
Fixed with https://github.com/trellis-ldp/trellis/commit/eb1add15dfcf61e862d4a285d04f52ddb164250d

PR #416, opened 2019-06-13
Switching to make no-op impls the CDI @Alternatives

ajs6f:
Alternative to https://github.com/trellis-ldp/trellis/pull/413

Based on [the CDI spec](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#builtin_qualifiers), leaving the no-op impls as `@Alternative`s and the others as (un-annotated) `@Default`s, we should arrive at the behavior that having any non-no-op impl (e.g. `WebACService` or `JmsPublisher`) on the classpath in a JAR with a `beans.xml` (presumably from build) should prefer the substantive impl, which, IIUC, meets @gregjan's immediate needs.

### DIFF
--- a/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
@@ -44,7 +44,6 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
-import javax.enterprise.inject.Alternative;
 import javax.inject.Inject;
 
 import org.apache.commons.rdf.api.Graph;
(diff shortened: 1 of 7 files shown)


ajs6f:
Also fixes https://github.com/trellis-ldp/trellis/issues/400

ajs6f:
@acoburn I'd like to let this hang until I can supply an accompanying PR for `trellis-cassandra` and @gregjan can decide whether this direction suits him.

ajs6f:
@acoburn, I just ran into this again working on the MicroProfile app. Can we merge this now? If @gregjan runs into problms because of it, they cannot be any worse than what he was having and I will be ready to help him if needed. I'm blocked (because of a conflict between `NoopNamespaceService` and `DBNamespaceService`) without this.

ajs6f:
Oops-- just noticed that you already approved this. I'll merge and move on.

PR #424, opened 2019-07-02
Added no-arg constructors to support CDI proxies.

gregjan:
These fixes, plus some adjustments on the trellis-cassandra side, have resulted in a working repository with basic auth and webac.

### DIFF
--- a/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
@@ -96,13 +96,10 @@ public class WebACService implements AccessControlService {
     private final boolean checkMembershipResources;
 
     /**
-     * No-op for CDI. Create a WebAC-based authorization service.
+     * Create a WebAC-based authorization service.
      */
     public WebACService() {
-        //this(getDefaultResourceService());
-    	this.resourceService = null;
-    	this.cache = null;
-    	this.checkMembershipResources = Boolean.FALSE;
+        this(getDefaultResourceService());
     }
 
     /**
@@ -136,8 +133,8 @@ public class WebACService implements AccessControlService {
      */
     public WebACService(final ResourceService resourceService,
             final CacheService<String, Set<IRI>> cache, final boolean checkMembershipResources) {
-        this.resourceService = requireNonNull(resourceService, "A non-null ResourceService must be provided!");
-        this.cache = requireNonNull(cache, "A non-null Cache must be provided!");
+        this.resourceService = resourceService;
+        this.cache = cache;
         this.checkMembershipResources = checkMembershipResources;
     }
 
--- a/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/WebACService.java
@@ -96,10 +96,13 @@ public class WebACService implements AccessControlService {
     private final boolean checkMembershipResources;
 
     /**
-     * Create a WebAC-based authorization service.
+     * No-op for CDI. Create a WebAC-based authorization service.
      */
     public WebACService() {
-        this(getDefaultResourceService());
+        //this(getDefaultResourceService());
+    	this.resourceService = null;
+    	this.cache = null;
+    	this.checkMembershipResources = Boolean.FALSE;
     }
 
     /**
--- a/core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
+++ b/core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
@@ -63,6 +63,14 @@ public class AgentAuthorizationFilter implements ContainerRequestFilter {
     private final Set<String> adminUsers;
 
     /**
+     * No-op constructor for CDI.
+     */
+    AgentAuthorizationFilter() {
+        this.agentService = null;
+        this.adminUsers = null;
+    }
+    
+    /**
      * Create an authorization filter.
      *
      * @param agentService the agent service
--- a/core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
+++ b/core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
@@ -69,7 +69,7 @@ public class AgentAuthorizationFilter implements ContainerRequestFilter {
         this.agentService = null;
         this.adminUsers = null;
     }
-    
+
     /**
      * Create an authorization filter.
      *
--- a/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
+++ b/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
@@ -89,16 +89,14 @@ public class WebAcFilter implements ContainerRequestFilter, ContainerResponseFil
     private final String baseUrl;
 
     /**
-     * No-op for CDI
-     *
-     * @param accessService the access service
+     * No-op constructor for CDI.
      */
     WebAcFilter() {
-    	this.accessService = null;
-    	this.challenges = null;
-    	this.baseUrl = null;
+        this.accessService = null;
+        this.challenges = null;
+        this.baseUrl = null;
     }
-    
+
     /**
      * Create a new WebAc-based auth filter.
      *
--- a/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
+++ b/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
@@ -89,6 +89,17 @@ public class WebAcFilter implements ContainerRequestFilter, ContainerResponseFil
     private final String baseUrl;
 
     /**
+     * No-op for CDI
+     *
+     * @param accessService the access service
+     */
+    WebAcFilter() {
+    	this.accessService = null;
+    	this.challenges = null;
+    	this.baseUrl = null;
+    }
+    
+    /**
      * Create a new WebAc-based auth filter.
      *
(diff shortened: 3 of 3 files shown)


ajs6f:
LUV THAT CHECKSTYLE

```
> Task :trellis-http:checkstyleMain FAILED
[ant:checkstyle] [ERROR] /home/travis/build/trellis-ldp/trellis/core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java:72: Trailing whitespace [RegexpSingleline]
[ant:checkstyle] [ERROR] /home/travis/build/trellis-ldp/trellis/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java:91: First sentence should end with a period. [JavadocStyle]
[ant:checkstyle] [ERROR] /home/travis/build/trellis-ldp/trellis/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java:97:5: Line contains a tab character. [FileTabCharacter]
[ant:checkstyle] [ERROR] /home/travis/build/trellis-ldp/trellis/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java:98:5: Line contains a tab character. [FileTabCharacter]
[ant:checkstyle] [ERROR] /home/travis/build/trellis-ldp/trellis/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java:99:5: Line contains a tab character. [FileTabCharacter]
[ant:checkstyle] [ERROR] /home/travis/build/trellis-ldp/trellis/core/http/src/main/java/org/trellisldp/http/WebAcFilter.java:101: Trailing whitespace [RegexpSingleline]
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':trellis-http:checkstyleMain'.
> Checkstyle rule violations were found. See the report at: file:///home/travis/build/trellis-ldp/trellis/core/http/build/reports/checkstyle/main.html
  Checkstyle files with violations: 2
  Checkstyle violations by severity: [error:6]
```

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
```diff
@@ -96,10 +96,13 @@
     private final boolean checkMembershipResources;
 
     /**
-     * Create a WebAC-based authorization service.
+     * No-op for CDI. Create a WebAC-based authorization service.
      */
     public WebACService() {
-        this(getDefaultResourceService());
+        //this(getDefaultResourceService());
+    	this.resourceService = null;
```
ajs6f: Is this actually a safe way to build a `WebACService`? Will it work if this ctor is used?

core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
```diff
@@ -62,6 +62,14 @@
     private final AgentService agentService;
     private final Set<String> adminUsers;
 
+    /**
+     * No-op constructor for CDI.
+     */
+    AgentAuthorizationFilter() {
+        this.agentService = null;
```
ajs6f: Ditto comment above.

core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
```diff
@@ -88,6 +88,17 @@
     private final List<String> challenges;
     private final String baseUrl;
 
+    /**
+     * No-op for CDI
+     *
+     * @param accessService the access service
+     */
+    WebAcFilter() {
+    	this.accessService = null;
```
ajs6f: Ditto.

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
acoburn: It would be better to leave this line as it is: `this(getDefaultResourceService());`

core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
acoburn: This constructor is actually fine -- it's package-protected in scope, so no one will be able to invoke it outside of the `org.trellisldp.http` package.

core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
acoburn: This is also fine, but you'll need to adjust the javadoc above, removing the `@param accessService`

acoburn:
The checkstyle rules are just so we don't have to argue about tabs-vs-spaces :-)

Though in fact, you will need to expand those tabs.

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
acoburn: I would remove all of the changes to this constructor.

core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
ajs6f: I disagree-- I think we should not a have a ctor that doesn't actually work in the codebase, period. It is confusion waiting to happen down the line.

core/http/src/main/java/org/trellisldp/http/WebAcFilter.java
ajs6f: Same as above-- I don't think it's okay to leave actual broken code in place if "no one can see it". Why couldn't we use a `ServiceLocator`-based approach as we do elsewhere?

core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
acoburn: What about this [commit](https://github.com/trellis-ldp/trellis/commit/f8612bc63a1574b4e3fe1638b33ab3bd95a5b38f)? We are adding these sorts of constructors elsewhere. Why not here, too?

gregjan:
If you leave it, then it will throw a NPE in a CDI environment, when it
uses the no-arg constructor to make a proxy.

core/http/src/main/java/org/trellisldp/http/AgentAuthorizationFilter.java
ajs6f: OUCH! I was wrong then-- I got into too much of a hurry. To me _this_ exactly where the no-op pattern _does_ makes sense.

acoburn:
Instead of changing this ctor, I'd rather have you remove the `requireNonNull` check on line 136 or refactor the `getDefaultResourceService` method so that it doesn't return `null` (line 290).

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
gregjan: If I do that, then I will have to remove the null checks, since getDefaultResourceService() returns a null in the CDI proxy context.

ajs6f:
Just to be clear, the NPEs are coming from explicit `null`-checks on parameters, not from failing business logic that assumes the presence of a real object in some reference? If so, scrap 'em. I'm a big fan of fail-fast for this stuff, but that's _too_ fast to integrate easily with CDI.

gregjan:
I'm taking another swing at this one. (not that swing)

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
```diff
@@ -133,7 +133,7 @@ public WebACService(final ResourceService resourceService,
      */
     public WebACService(final ResourceService resourceService,
             final CacheService<String, Set<IRI>> cache, final boolean checkMembershipResources) {
-        this.resourceService = requireNonNull(resourceService, "A non-null ResourceService must be provided!");
+        this.resourceService = resourceService;
         this.cache = requireNonNull(cache, "A non-null Cache must be provided!");
```
ajs6f: I would take it out here too, or we'll just have to do it later.

ajs6f:
This is where the tradeoffs lead us. 👍

components/webac/src/main/java/org/trellisldp/webac/WebACService.java
gregjan: It's gone from there.

ajs6f: Sorry, I meant `this.cache = requireNonNull(cache, "A non-null Cache must be provided!");`. Or is that not what you meant?

gregjan: Ah, okay. I'll take out that one too.

acoburn:
Maybe some of this would be easier if we had a no-op `ResourceService` class (it would always return a `MISSING_RESOURCE`)

ajs6f:
That's what I meant [by this comment](https://github.com/trellis-ldp/trellis/pull/424/commits/ef2c2b2cc75a7991bda801cc92a02d9896751567#r299627419). Frankly, once this gets merged I was going to just dash in and add no-op impls anywhere they could appear, to make myself feel better about the whole "ctors that build non-functional objects" scruple I expressed.

acoburn:
@ajs6f that sounds like an excellent plan.

ajs6f:
I think we're all in agreement-- what @gregjan has here is good, and once merged, I'll polish it a bit in ways that shouldn't cause any benefit or harm to CDI users (which now include @gregjan and in future should be most people).

acoburn:
@ajs6f if you're happy with the state of this PR, feel free to merge it. Then we can move forward with those no-op ctors.

PR #426, opened 2019-07-03
Add some tests to the no-op ctors

acoburn:


core/http/src/test/java/org/trellisldp/http/AgentAuthorizationFilterTest.java
```diff
@@ -81,4 +81,13 @@ public void testFilterAdminAgent() throws Exception {
         verify(mockContext).setProperty(eq(SESSION_PROPERTY), sessionArgument.capture());
         assertEquals(Trellis.AnonymousAgent, sessionArgument.getValue().getAgent(), "Unexpected agent IRI!");
     }
+
+    @Test
+    public void testNoopCtor() throws Exception {
```
ajs6f: This seems like a slightly confusing name. Maybe `testNoParamCtor`?

core/http/src/test/java/org/trellisldp/http/WebAcFilterTest.java
```diff
@@ -186,4 +186,9 @@ public void testFilterResponseBaseUrl() throws Exception {
         assertEquals("acl", link.getRel());
         assertEquals("http://example.com/?ext=acl", link.getUri().toString());
     }
+
+    @Test
+    public void testNoopFilterCtor() {
```
ajs6f: Similar, maybe `testNoParamCtor`?

core/http/src/test/java/org/trellisldp/http/WebAcFilterTest.java
```diff
@@ -186,4 +186,9 @@ public void testFilterResponseBaseUrl() throws Exception {
...
+        assertDoesNotThrow(() -> new WebAcFilter());
```
ajs6f: Or `         assertDoesNotThrow(WebAcFilter::new);`? You know I have a serious fetish for point-free style. :grin:

acoburn: Actually, that's what I first wrote, and the compiler didn't like it, so I had to use the longer form.

ajs6f: Very interesting (although obviously not important). I might look into that after you merge this. I wonder why?

acoburn: My suspicion is that the `assertDoesNotThrow` [method](https://junit.org/junit5/docs/5.3.0/api/org/junit/jupiter/api/Assertions.html) accepts two types of arguments, both of which are functional interfaces, and it can't decide between an `Executor` and a `ThrowingSupplier<T>`.

PR #428, opened 2019-07-05
Move all WebAC-related code into trellis-webac

### LINKED ISSUE #417: Move all WebAC code into its own module
At present, WebAC authorization is split between the WebAC authorization implementation (`trellis-webac`) and the HTTP request filter (`http.WebAcFilter`). By consolidating all of the WebAC-related code in the `trellis-webac` module, we should be able to eliminate the `api.AccessControlService`, since that service will be more self-contained. The only possible sticking point concerns the `trellis-webdav` module and how WebAC gets applied to those WebDAV-specific methods.

See also the discussion [here](https://github.com/trellis-ldp/trellis/issues/209#issuecomment-501832349)

acoburn:
Resolves #417


acoburn:
This removes the `AccessControlService` interface from the core API, moves the JAX-RS request filter from the `http` module to the `webac` module, and removes the redundant WebDAV-based authZ filter, since we can create a loose-coupling of WebDAV and WebAC via configuration (rather than via Java dependencies)

acoburn:
Another note on this PR: in order to avoid a hard dependency between WebDAV and WebAC, there is some additional configuration that needs to be added to a WebDAV-enabled server that uses WebAC. In particular, the following properties need to be configured:

```
trellis.webac.method.readable=PROPFIND
trellis.webac.method.writable=PROPPATCH,COPY,MOVE
trellis.webac.method.appendable=MKCOL
```

ajs6f:
This basically makes sense to me. We can reintroduce an `AccessControlService` type if need be. Is there some documentation about the new configuration somewhere? It's not quite obvious to me.

acoburn:
I added some documentation in the most recent commit. Eventually, this should also go onto the wiki somewhere.

PR #429, opened 2019-07-09
Add authZ support for acl:origin

### LINKED ISSUE #9: Add support for acl:origin
https://www.w3.org/wiki/WebAccessControl#Cors_User_Agents

acoburn:
Resolves #9

The logic here assumes that any `acl:origin` defined in an ACL has already been white-listed at the server level. That is, setting `<> acl:origin <https://bad.example.com>` in an ACL document does not suddenly give that Origin access to a resource if the server configuration hasn't already given that Origin access (e.g. via `*` or specifically by name)

### DIFF
--- a/components/webac/src/main/java/org/trellisldp/webac/Authorization.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/Authorization.java
@@ -13,10 +13,15 @@
  */
 package org.trellisldp.webac;
 
-import static java.util.Collections.unmodifiableSet;
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptySet;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.collectingAndThen;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.mapping;
+import static java.util.stream.Collectors.toSet;
 
-import java.util.HashMap;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


components/webac/src/main/java/org/trellisldp/webac/Authorization.java
```diff
@@ -142,4 +143,13 @@ public BlankNodeOrIRI getIdentifier() {
     public Set<IRI> getDefault() {
         return unmodifiableSet(dataMap.get(ACL.default_));
     }
+
+    /**
+     * Retrieve the acceptable origins of the ACL document.
+     *
+     * @return the origin IRIs
+     */
+    public Set<IRI> getOrigin() {
```
ajs6f: If this gets called more than once and since `dataMap` doesn't seem to change it might be worth building this `Set<IRI>` and keeping it in a field to avoid building short-lived objects.

components/webac/src/main/java/org/trellisldp/webac/Authorization.java
```diff
@@ -68,17 +77,11 @@ public Authorization(final BlankNodeOrIRI identifier, final Graph graph) {
...
-        this.dataMap.put(ACL.accessTo, new HashSet<>());
-        this.dataMap.put(ACL.default_, new HashSet<>());
-        this.dataMap.put(ACL.origin, new HashSet<>());
-
-        graph.stream(identifier, null, null).filter(triple -> dataMap.containsKey(triple.getPredicate()))
+        this.dataMap = graph.stream(identifier, null, null)
+            .filter(triple -> predicates.contains(triple.getPredicate()))
             .filter(triple -> triple.getObject() instanceof IRI)
-            .forEachOrdered(triple -> dataMap.get(triple.getPredicate()).add((IRI) triple.getObject()));
+            .collect(groupingBy(Triple::getPredicate, mapping(triple -> (IRI) triple.getObject(),
```
ajs6f: Fancy!

components/webac/src/main/java/org/trellisldp/webac/Authorization.java
acoburn: That's an excellent point. I reworked the constructor to `.collect` the values directly into immutable `Set<IRI>` objects so that there is no need to create new objects.

PR #433, opened 2019-07-18
Accept-Post should include the wildcard type

### LINKED ISSUE #432: Accept-POST should include the wildcard type
The `Accept-Post` response header currently lists the RDF serializations that are supported by Trellis, which is correct. But since Trellis also supports NonRDFSource types, the `Accept-Post` header should _also_ include the wildcard type (`*/*`)

acoburn:
Resolves #432

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/impl/GetHandler.java
+++ b/core/http/src/main/java/org/trellisldp/http/impl/GetHandler.java
@@ -389,8 +389,10 @@ public class GetHandler extends BaseLdpHandler {
             builder.link(type.getIRIString(), "type");
             // Mementos don't accept POST or PATCH
             if (LDP.Container.equals(type) && !isMemento) {
-                builder.header(ACCEPT_POST, concat(getServices().getIOService().supportedWriteSyntaxes().stream()
-                        .map(RDFSyntax::mediaType), of(WILDCARD)).collect(joining(",")));
+                final List<RDFSyntax> rdfSyntaxes = getServices().getIOService().supportedWriteSyntaxes();
+                final Stream<String> allSyntaxes = concat(rdfSyntaxes.stream().map(RDFSyntax::mediaType), of(WILDCARD));
+
+                builder.header(ACCEPT_POST, allSyntaxes.collect(joining(",")));
             } else if (LDP.Resource.equals(type) && !isMemento) {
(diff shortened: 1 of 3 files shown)


core/http/src/main/java/org/trellisldp/http/impl/OptionsHandler.java
```diff
@@ -124,8 +127,8 @@ public ResponseBuilder ldpOptions(final ResponseBuilder builder) {
             } else {
                 // Containers support POST
                 builder.header(ALLOW, join(",", GET, HEAD, OPTIONS, PATCH, PUT, DELETE, POST));
-                builder.header(ACCEPT_POST, getServices().getIOService().supportedWriteSyntaxes().stream()
-                        .map(RDFSyntax::mediaType).collect(joining(",")));
+                builder.header(ACCEPT_POST, concat(getServices().getIOService().supportedWriteSyntaxes().stream()
```
ajs6f: IIUC, you're concating `WILDCARD` onto a stream from `getServices().getIOService().supportedWriteSyntaxes()`, but this is getting a bit hard to read. How about:
```
rdfSyntaxes = getServices().getIOService().supportedWriteSyntaxes();
allSyntaxes = concat(rdfSyntaxes.stream().map(RDFSyntax::mediaType), of(WILDCARD));
builder.header(ACCEPT_POST, allSyntaxes.collect(joining(",")));
```
modulo appropriate type declarations, or something a little more broken-out?

core/http/src/main/java/org/trellisldp/http/impl/GetHandler.java
```diff
@@ -387,8 +389,8 @@ private void addLdpHeaders(final ResponseBuilder builder, final IRI model) {
             builder.link(type.getIRIString(), "type");
             // Mementos don't accept POST or PATCH
             if (LDP.Container.equals(type) && !isMemento) {
-                builder.header(ACCEPT_POST, getServices().getIOService().supportedWriteSyntaxes().stream()
-                        .map(RDFSyntax::mediaType).collect(joining(",")));
+                builder.header(ACCEPT_POST, concat(getServices().getIOService().supportedWriteSyntaxes().stream()
```
ajs6f: See below.

core/http/src/main/java/org/trellisldp/http/impl/OptionsHandler.java
acoburn: Yes, that would make it more readable.

Issue #358, opened 2019-02-12
Consider replacing Commons-RDF with Jena types


acoburn:
Commons-RDF provides some nice abstractions, especially when working across multiple RDF implementations, but in practice, this seems to just add work for the garbage collector.

If we just use Jena types directly, the number of small objects will be reduced significantly. It would also mean that we could simplify some of the `*Service` interfaces: for example, the features of `IOService` (and a few others) can go directly into the core HTTP layer.

ajs6f:
👍

acoburn:
I've started looking into this, and I'd like a little feedback on some of the type changes that will be involved.

Wherever we have a CommonsRDF `Graph`, that can simply become a Jena `Model`.
Likewise, a CommonsRDF `Dataset` can become a Jena `DatasetGraph`.

In cases where we stream out CommonsRDF `Quad` or `Triple` objects, I think these should be Jena `Quad` and `Triple` objects (rather than `Statement`, since they are not connected to a particular `Model`)

Most of the remaining interface changes would involve changing CommonsRDF `IRI` to a Jena `Resource`

There are a few instances where there is a `Model`-free mapping from a BlankNode to an IRI; here, I think it might make sense to do a Jena `Node` to `Node` conversion, since these are not tied to a particular `Model`.

Otherwise, if we start using the Jena `Resource` interface, we may want to change the Trellis `Resource` to something else. How about `TrellisResource`?

pmcb55:
Hi,

Just a quick comment on this - did you ever consider moving to RDF4J (formally known as Sesame) instead of continuing with Jena? Someone elsewhere on this repo mentioned something about the '**_legacy inertia of Jena_**' or something, and in my opinion RDF4J is a far more commercial, enterprise-grade library, e.g. it forms the basis of GraphDB from Ontotext, and what was BlazeGraph (now Amazon Neptune). It's also very actively and professionally maintained, and is laser focused on being a pure open-source Java library for RDF and SPARQL (rather than being a mismash of research-related extensions for things like OWL!).

I'm just playing devil's advocate here, so please take my tongue-in-cheek comparison of Jena vs RDF4J above with a very large grain of salt :) !

Cheers,

Pat.

acoburn:
[...]

PR #435, opened 2019-07-21
Make ETag generation more flexible

### LINKED ISSUE #434: Make ETag generation more flexible
Different implementations may have different ideas/strategies for generating ETags. By creating an interface in the HTTP layer with default method implementations, this can be made considerably more flexible.

acoburn:
Resolves #434

This also simplifies the default etag generation code, making it, by default, invariant across response content-types and/or prefer headers, which also makes it considerably more usable by clients.

In addition, should an implementation choose to implement a different mechanism for generating ETags, that is easily done via the standard `ServiceLoader` mechanism.

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/core/EtagGenerator.java
+++ b/core/http/src/main/java/org/trellisldp/http/core/EtagGenerator.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.trellisldp.http.core;
+
+import static org.apache.commons.codec.digest.DigestUtils.md5Hex;
+
+import org.trellisldp.api.Resource;
+
+public interface EtagGenerator {
+
+    /**
+     * Generate the value portion for an entity tag for the given resource.
+     * @param resource the Trellis resource
+     * @return the value for the entity tag
+     */
+    default String getValue(final Resource resource) {
+        return md5Hex(resource.getModified().getNano() + "." + resource.getIdentifier());
+    }
+}
--- a/core/http/src/main/java/org/trellisldp/http/core/EtagGenerator.java
+++ b/core/http/src/main/java/org/trellisldp/http/core/EtagGenerator.java
@@ -21,6 +21,10 @@ public interface EtagGenerator {
 
     /**
      * Generate the value portion for an entity tag for the given resource.
+     *
+     * @implSpec The default implementation of this method generates an ETag value based on
+     *           the precision of {@link Resource#getModified} implementation, up to and
+     *           including nano-second precision.
      * @param resource the Trellis resource
      * @return the value for the entity tag
      */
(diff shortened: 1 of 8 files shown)


core/http/src/main/java/org/trellisldp/http/core/EtagGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+
+public interface EtagGenerator {
+
+    /**
+     * Generate the value portion for an entity tag for the given resource.
+     * @param resource the Trellis resource
+     * @return the value for the entity tag
+     */
+    default String getValue(final Resource resource) {
+        return md5Hex(resource.getModified().getNano() + "." + resource.getIdentifier());
```
ajs6f: Do we need to document that while ETags are only good up to a second, we suggest (or require) that the value includes more?

core/http/src/main/java/org/trellisldp/http/impl/HttpUtils.java
```diff
@@ -393,6 +382,15 @@ public static void closeDataset(final Dataset dataset) {
         }
     }
 
+    /**
+     * Load an Etag generator.
+     * @return the etag generator
+     */
+    public static EtagGenerator loadEtagGenerator() {
```
ajs6f: Do we want to bake this into the framework or should we `@Inject` a generator where needed?

core/http/src/main/java/org/trellisldp/http/core/EtagGenerator.java
acoburn: ETags can have an arbitrary level of precision. It's the `If-Modified-Since`-category of conditional requests that have a limitation on their precision (i.e. one second). In this case, the precision of the ETag will be based on the level of precision of the underlying `Resource` implementation. Adding that in a note to the method would make sense.

ajs6f: Of course you're completely right-- I need more tea this morning before looking at code.

core/http/src/main/java/org/trellisldp/http/impl/HttpUtils.java
acoburn: I don't have a strong opinion -- I basically assume that everyone will use this (default) generator. Adding a different one would simply be a matter of registering it with the service loader, which seems a little lighter-weight than using `@Inject`. Or do you think `@Inject` would be better here?

Part of my logic for using the `ServiceLoader` related to how this gets loaded from "the inside" rather than being registered like the `ResourceService` or `EventService` via the `ServiceBundler`.

acoburn: Also, `@Inject` would require an actual implementation of the interface, while here, we have an anonymous implementation that is created only if another impl isn't available.

ajs6f: I'm not too worried about the cost of a single object that lives for the entire application lifetime, and I think the cost of having two different canonical means by which services are wired is much, much higher. If we want to inject, let injection inject.

acoburn: At present, we are using the `ServiceLoader` for the following:

* All Commons-RDF injection (to be removed in #358)
* Injection of `ConstraintRules` in the [http-layer](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/impl/BaseLdpHandler.java#L46)
* Injection of `TimemapGenerator` in the [http-layer](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/impl/MementoResource.java#L331-L334)
* Injection of `IdentifierService` (throughout)
* Injection of `ActivityStreamService` (in the notification implementations)

Everything else uses plain-old `@Inject`

I can't say that there is a clearly defined distinction between these two categories other than the ServiceLoader-based services tend to be "simpler" (given some definition of "simple").

Perhaps we should merge this change now and address the larger issue of `@Inject` vs. `ServiceLoader` elsewhere?

ajs6f: I'm fine with that-- can you put the above list in a ticket and include this new addition?

ajs6f: https://github.com/trellis-ldp/trellis/issues/441

PR #439, opened 2019-07-21
Add a no-op resource service implementation

### LINKED ISSUE #437: Implement a No-op ResourceService
This will be useful for CDI

acoburn:
Resolves #437

### DIFF
--- a/core/api/src/main/java/org/trellisldp/api/NoopResourceService.java
+++ b/core/api/src/main/java/org/trellisldp/api/NoopResourceService.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.trellisldp.api;
+
+import static java.util.Collections.emptySet;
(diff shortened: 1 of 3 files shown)


core/api/src/main/java/org/trellisldp/api/NoopResourceService.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Set;
+import java.util.concurrent.CompletionStage;
+
+import org.apache.commons.rdf.api.Dataset;
+import org.apache.commons.rdf.api.IRI;
+
+/**
+ * A no-op resource service that can be used with CDI and proxy objects.
+ */
```
ajs6f: I think we'd want to make this a CDI `@Alternative` like the other no-op-sters, no?

core/api/src/main/java/org/trellisldp/api/NoopResourceService.java
```diff
@@ -0,0 +1,66 @@
...
+public class NoopResourceService implements ResourceService {
+
+    @Override
+    public CompletionStage<Resource> get(final IRI identifier) {
+        return completedFuture(MISSING_RESOURCE);
+    }
+
+    @Override
+    public CompletionStage<Void> replace(final Metadata metadata, final Dataset dataset) {
+        return completedFuture(null);
```
ajs6f: Getting a bit prissy, but maybe one constant `private final static CompletionStage<Void> NO_RESULT = completedFuture(null)` to be used over and over instead of creating new objects? Yeah, it's a lazy hot Sunday which is why I'm being weird.

core/api/src/main/java/org/trellisldp/api/NoopResourceService.java
acoburn: Yes, definitely.

core/api/src/main/java/org/trellisldp/api/NoopResourceService.java
acoburn: This has been added in the latest commit.

PR #443, opened 2019-07-22
Relocate ServiceBundler interface to HTTP layer

### LINKED ISSUE #442: Relocate ServiceBundler from api to http layer
All uses of the ServiceBundler are at the HTTP layer or higher, and in order to use the interfaces defined at the HTTP layer (e.g. `TimemapGenerator`, `EtagGenerator`) we would either need to bring those interface into the API layer or move the ServiceBundler into the HTTP layer.

I propose moving `ServiceBundler` into the `org.trellisldp.http.core` package.

This would be a prerequisite to working on #441

acoburn:
Resolves #442


acoburn:
Yes, I'll add those changes in a separate PR.

Issue #441, opened 2019-07-22
Review service injection mechanisms


acoburn:
At present, there are two principle mechanisms in use for service injection: `ServiceLoader` and `@Inject`. We should discuss how services should be injected throughout the code in order to make this work more consistently. See [this discussion](https://github.com/trellis-ldp/trellis/pull/435#discussion_r305872716) for more context.

The following services are currently injected via the `ServiceLoader`:

* All Commons-RDF injection (to be removed in #358)
* Injection of ConstraintRules in the [http-layer](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/impl/BaseLdpHandler.java#L46) -- we should be able to support multiple service implementations here
* Injection of TimemapGenerator in the [http-layer](https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/impl/MementoResource.java#L331-L334) -- here we use an anonymous impl of this interface
* Injection of ETagGenerator in the http-layer (see #435)
* Injection of IdentifierService (throughout)
* Injection of ActivityStreamService (in the notification implementations)

`@Inject` is used elsewhere, but there is no clear way we are distinguishing between these two categories (though for Commons-RDF using the `ServiceLoader` is the preferred mechanism)

ajs6f:
It seems to me that another way to ask this question is: are we willing to commit to CDI throughout Trellis? If yes, we should use `@Inject` everywhere because there is no advantage to `ServiceLoader` then. If not, we should not rely on `@Inject` at all, so every injection point should feature some way to avoid using injection (the easy obvious way to is to use ctor injection all the time). I'm not sure what `ServiceLoader` adds in that case too, at least that we couldn't buy ready-built-- it seems like a way to have a default choice for a wiring slot that can be overridden, which is exactly the function of a wiring framework.

acoburn:
Thus far we have had success with constructor injection. This has the added advantage that everything also works in non-CDI environments (e.g. Dropwizard or Spring Boot). This also makes unit testing a lot easier.

In principle, I suspect that we could just add slots for most of these in the `ServiceBundler` interface. The `ActivityStreamService` interface could be added to the various notification constructors.

ajs6f:
That would work for me. It would also give us a nice place (`ServiceBundler`) to send people who want to use a new wiring framework.

acoburn:
[...]

PR #444, opened 2019-07-23
Move services into the ServiceBundler interface

acoburn:
Addresses #441

This moves many of the internally-loaded services into the `ServiceBundler` interface.

With the exception of Commons-RDF, all of the services used in `./core` are now defined in the `ServiceBundler`.

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/impl/BaseLdpHandler.java
+++ b/core/http/src/main/java/org/trellisldp/http/impl/BaseLdpHandler.java
@@ -20,14 +20,11 @@ import static javax.ws.rs.core.HttpHeaders.IF_UNMODIFIED_SINCE;
 import static org.trellisldp.api.TrellisUtils.getInstance;
 
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.List;
 
 import javax.ws.rs.core.EntityTag;
 
 import org.apache.commons.rdf.api.IRI;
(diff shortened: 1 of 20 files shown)


acoburn:
A note on the use of `Iterable<ConstraintService>`: this gives us flexibility to use this in both CDI and non-CDI environments. In a CDI environment, one would expect a `javax.enterprise.inject.Instance` in that slot, and the `Instance` has `Iterable` as a super interface, which is all we really need in `./core`. In that way, both a regular `List<ConstraintService>` and an `Instance<ConstraintService>` will work fine.

core/http/src/main/java/org/trellisldp/http/impl/BaseLdpHandler.java
```diff
@@ -66,6 +58,7 @@ protected BaseLdpHandler(final TrellisRequest request, final ServiceBundler serv
         this.requestBaseUrl = getRequestBaseUrl(request, baseUrl);
         this.request = request;
         this.services = services;
+        services.getConstraintServices().forEach(constraintServices::add);
```
ajs6f: Can we not just take a reference to the constraint services and use them from that reference?

acoburn: Definitely. I just updated the PR to do just that.

Issue #446, opened 2019-07-26
Stereotype for no-op impls


ajs6f:
We could be more precise in our code by using CDI stereotypes as described [here](https://github.com/trellis-ldp/trellis/pull/445#discussion_r307922503).

PR #447, opened 2019-07-28
Use CDI stereotype for concision

ajs6f:
https://github.com/trellis-ldp/trellis/issues/446


ajs6f:
Ach- forgot to run `check`. Thanks for the assist!

PR #448, opened 2019-07-31
Update injection point for IOService

acoburn:
This simplifies the IOService injection in a CDI context.

### DIFF
--- a/components/io-jena/src/main/java/org/trellisldp/io/JenaIOService.java
+++ b/components/io-jena/src/main/java/org/trellisldp/io/JenaIOService.java
@@ -122,6 +122,7 @@ public class JenaIOService implements IOService {
     /**
      * Create a serialization service.
      */
+    @Inject
     public JenaIOService() {
         this(new NoopNamespaceService());
     }
@@ -151,7 +152,6 @@ public class JenaIOService implements IOService {
      * @param htmlSerializer the HTML serializer service
      * @param cache a cache for custom JSON-LD profile resolution
      */
-    @Inject
     public JenaIOService(final NamespaceService namespaceService,
             final RDFaWriterService htmlSerializer,
             @TrellisProfileCache final CacheService<String, String> cache) {


acoburn:
We previously had the injection point for `JenaIOService` defined for the no-arg constructor, but when I [refactored](https://github.com/trellis-ldp/trellis/commit/d26f25e05f3372705e50da894e457d227609833b#diff-02193fd35d259b5bb39af2d29e29f010L128) out our use of the ServiceLoader, I adjusted the injection point for this bean.

Then, when @jholleran and I were trying to bring the `trellis-cassandra` code to be current with the Trellis API (see [this branch](https://github.com/jholleran/trellis-cassandra/tree/feature/update-to-latest-service-bundler-interface) -- PR will be forthcoming, once this is merged), this was one of the changes we needed to make to get injection working properly with the `trellis-cassandra` code.

In principle, I agree that this change shouldn't really be needed, and that we should be able to wire up a complete CDI application as-is, but in practice, I wasn't able to get this to work. Once we have a CDI-based application set up in the main Trellis repo, I think we can settle on a clear set of injection points, but for now, there's a little bit of jitter in terms of getting CDI to work in the downstream trellis-cassandra code.

Does that provide sufficient background?

ajs6f:
I'm afraid not-- the crux is in the statement
> this was one of the changes we needed to make to get injection working properly with the trellis-cassandra code.
Why? What exactly did you see? It would be valuable to record that, if nothing else. Again, I have no objection to this change, I just want to understand it thoroughly. For one thing, someone who wants to use a functional (not no-op) `NamespaceService` with CDI is going to find that much harder after this change, so it would be good to be able to explain very clearly why.

acoburn:
The issue had more to do with the `RDFaWriterService`, which is a nullable arg to the `JenaIOService`, and that bean wasn't getting wired in properly. The specific error was:

```
Caused by: org.jboss.weld.exceptions.DeploymentException: WELD-001408: Unsatisfied dependencies for type NamespaceService with qualifiers @Default
 at injection point [BackedAnnotatedParameter] Parameter 1 of [BackedAnnotatedConstructor] @Inject public org.trellisldp.rdfa.HtmlSerializer(NamespaceService)
 at org.trellisldp.rdfa.HtmlSerializer.<init>(HtmlSerializer.java:0)
```

That is, the core issue is getting the `HtmlSerializer` working correctly in CDI (it currently doesn't). So the bigger question is perhaps: Should we address the larger issue now? or shall we make a small adjustment now in order to get the trellis-cassandra code buildable (it doesn't currently build/test successfully)

ajs6f:
I'm confused-- the problem is reported as `Unsatisfied dependencies for type NamespaceService with qualifiers @Default at injection point ... @Inject public org.trellisldp.rdfa.HtmlSerializer(NamespaceService)`. The problem is that CDI can't find a `NamespaceService`-- that's what we need to fix. Did you alter [`beans.xml`](https://github.com/trellis-ldp/trellis-cassandra/blob/master/webapp/src/main/resources/META-INF/beans.xml) in some way? The listing of `NoopNamespaceService` there under `alternatives` should be enough to have one available.

acoburn:
Interesting. I'll experiment a little more now that some of these other PRs have come through.

ajs6f:
Cool- hook me back in if/when needed! I would first make sure `beans.xml` is doing what you want it to do-- in situations like this, when you can be confident that a impl of the requisite type is available but a bean isn't getting built from it, that's often the problem.

acoburn:
I am able to get `trellis-cassandra` to work without this change. See [this PR](https://github.com/trellis-ldp/trellis-cassandra/pull/68), which still encounters an issue with the `NamespaceService` not being injected, but there is a work-around. I would have expected the `beans.xml` configuration to pick up the `NoopNamespaceService` in the `<alternatives>` section, but that doesn't seem to play out in practice.

ajs6f:
That PR having been merged, I propose to close this one, if we don't need it. I am working on getting that `NamespaceService` oddity in `t-c*` straightened out (once I get Travis-CI straightened out arg!).

Issue #440, opened 2019-07-21
In-memory `ResourceService` for speedy and simple testing


ajs6f:
It might be nice to have an in-memory `ResourceService` for testing purposes. The advantage is less speed of execution and more simplicity, since testing should always use the simplest possible setup.

acoburn:
This could also replace a lot of the mocking code we currently have.

PR #450, opened 2019-08-01
Introducing in-memory services

ajs6f:
https://github.com/trellis-ldp/trellis/issues/440

First draft, needs tests

NOT YET FOR MERGING

### DIFF
--- a/components/test/src/main/java/org/trellisldp/test/InMemoryBinaryService.java
+++ b/components/test/src/main/java/org/trellisldp/test/InMemoryBinaryService.java
@@ -51,23 +51,23 @@ public class InMemoryBinaryService implements BinaryService {
     private final Map<IRI, Binary> data = new ConcurrentHashMap<>();
 
     @Override
-    public CompletionStage<? extends Binary> get(IRI identifier) {
+    public CompletionStage<? extends Binary> get(final IRI identifier) {
         return completedFuture(data.get(identifier));
     }
 
     @Override
-    public CompletionStage<Void> setContent(BinaryMetadata meta, InputStream bytes) {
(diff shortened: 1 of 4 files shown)


acoburn:
While this will be very useful for the test code, it _might_ be useful in other contexts, too. So, would it make sense to put these classes in their own `./component` jar? I.e. `./components/simple`.

I'd be happy to add the gradle parts if you'd like

ajs6f:
Yeah, I can see that making sense. Let me finish the basic work here (writing tests now) and we can send it wherever we want. I want to get back to the µprofile app-- just doing a little warm-up here since I haven't seen Trellis in quite a while!

ajs6f:
@acoburn This just needs some tests for the binary side. Do we have anything like `ResourceServiceTests` I could use for that? Otherwise, I have no freakin' idea what happened to Git there. Those look like commits in `master` _and_ duplicated commits in this branch? You have any idea how I could have done that?

acoburn:
maybe if you did a merge from the master branch? you can try rebasing off of master again. Alternatively, you could squash your commits and then cherry-pick them into a clean branch.

As for the binary tests, there's nothing stand-alone for that, but that's a good idea for something to add.

ajs6f:
I guess we can just squash this when it gets merged, right? 

I'll file a ticket for more tests, and meanwhile I'll add some simple tests here and be done.

ajs6f:
Okay, that's the last bit of testing for the mo, and rebasing cured the Git weirdity.

Let me know if you want more tests before merging. Otherwise, I'll wait to hear from you about [whether and where you want to move this](https://github.com/trellis-ldp/trellis/pull/450#issuecomment-517319417).

Issue #209, opened 2018-09-20
MicroProfile packaging and deployment


acoburn:
At present, there is a full-featured [Dropwizard](https://www.dropwizard.io)-based Trellis application along with a WAR-based application (sans any authN support) and a somewhat half-baked OSGi-based application. (The OSGi-based application is really meant to be a Proof-of-Concept app rather than a production-worthy application.)

In addition, it would be a step in the right direction to have a purely CDI-based application in the main Trellis repository using the Triplestore persistence layer. This would add a production-worthy, non-Dropwizard-based deployment option; it would also mean that we'd be testing CDI-based packaging with each build.

ajs6f:
We can generalize the Trellis-Cassandra CDI app for this, if that sounds interesting. To do that, we will have to once again deal with the connections between wiring and configuration, although we would have to do that in any case.

ajs6f:
Whether we do this starting from `trellis-cassandra` or not, we will need to think about wiring vs. config at least a little, but one way to avoid most of that discussion would be to commit to keeping the Dropwizard example and this example in sync, component-wise. Then they are both very much _examples_ and not _products_.

Would we be willing to use [Thorntail](https://thorntail.io/) for this? That's what I'm using in `t-c`, and besides shoddy logging because of a commitment they have to an obsolete JBoss logging library, it's fantastic.

acoburn:
I don't know much about Thorntail, but if you're happy with it, then I have no concerns at all about using that.

As for _examples_ vs. _products_, that is a very good way to view these. I see three levels of artifacts coming from the Trellis project:

1. Everything in `core`: including `api`, `http` and `vocabulary`. These are the true _products_ of Trellis and are the things I'm most concerned about getting "right".
2. All the component implementations, including what's in `components`, `notifications` and `auth` as well as the various `trellis-ext-*` repositories. These are concrete implementations of various parts of the SPI. I'm committed to those insofar as they are useful, but if the underlying tech becomes obsolete, then I'd just write a different implementation. These are, in a sense, "examples", though I think they're also (generally) robust examples.
3. The deployable artifacts. These are just concrete wirings of the service implementations. I'd like them to be robust, but again, they are just an assembly of the artifacts from 2, above.

ajs6f:
[...]

PR #453, opened 2019-08-01
Add default implementations for http.core interfaces

### LINKED ISSUE #449: Create default implementations for http.core interfaces
Two new interfaces in the http layer (`http.core.EtagGenerator` and `http.core.TimemapGenerator`) have no default implementations, which makes it harder to use these as injectable CDI beans, especially since they are now required parts of the `ServiceBundler` interface.

Related question: should these beans have a `@Default` and/or `@ApplicationScoped` annotation assigned?

acoburn:
Resolves #449

### DIFF
--- a/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisServiceBundler.java
+++ b/components/app-triplestore/src/main/java/org/trellisldp/app/triplestore/TrellisServiceBundler.java
@@ -40,6 +40,8 @@ import org.trellisldp.app.TrellisCache;
 import org.trellisldp.constraint.LdpConstraints;
 import org.trellisldp.file.FileBinaryService;
 import org.trellisldp.file.FileMementoService;
+import org.trellisldp.http.core.DefaultEtagGenerator;
+import org.trellisldp.http.core.DefaultTimemapGenerator;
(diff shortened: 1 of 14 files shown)


ajs6f:
I don't see any place these types are getting injected-- is that already in the code base?

acoburn:
At present, we just have Dropwizard, which doesn't do injection (c.f. #209). This PR does update the manual creation of these objects from anonymous `new EtagGenerator() { }` to `new DefaultEtagGenerator()`.

One consideration here is that the existing interfaces both have default methods, which constitute the entirety of the implementation  -- that code could legitimately be moved into these `Default*Generator` classes, leaving the interfaces much cleaner.

ajs6f:
No, we have the `trellis-cassandra` app and we have the µprofile app on the way, so we do need to think about how this stuff gets injected. In these particular cases, I think we should be okay if they appear in the `ServiceBundler` type in the usual way. (Required components always appear there by policy.) 

As for where to put the default behavior, I entirely agree that we should factor it over to the concrete default classes, especially since I took a look and the logic is not trivial. There is some real and serious business going on there.

acoburn:
In fact, the `trellis-cassandra` app is very much what prompted this PR.

I will refactor the impl code into the implementing classes. A new commit is on the way.

PR #460, opened 2019-08-08
Add ServiceBundler base class

acoburn:
Related to #209

While this PR does not add microprofile deployment support to Trellis, it will make adding such support much easier, and I am splitting this portion into a separate PR to make the review simpler.

In addition to adding a CDI module, this also separates the `trellis-triplestore` module from `trellis-audit`. At present, those two modules are tied together unnecessarily; pulling them apart makes CDI injection considerably easier and more generalizable.

This also changes the `javax.inject` dependency from the glassfish-derived dependency to the more standard `javax.inject` artifact.


ajs6f:
Separating out audit stuff LGTM, otherwise:

I'm not sure I understand the purpose of the CDI component. It doesn't create a deployable artifact.

If the part inside it to be reused is just the `CDIServiceBundler` impl, then I'll offer a different design with (I believe) more reuse potential. We can stick an `BaseServiceBundler` in `-app` that would look exactly like what you see in `CDIServiceBundler` modulo the injection of an `Instance`. (Keep the `javax.*` annotations for injection, because they only require the presence of the CDI API and do not affect operation at all if they are not used.) We would instead inject a new type
```
interface ConstraintServices extends Iterable<ConstraintService> {}
```
, so no client code changes, and no use of CDI-specific types in `-app`. A CDI app would have to impl `ConstraintServices` (probably via `Instance<ConstraintService>`, but that usage would be isolated in CDI instead of preventing reuse across code). `BaseServiceBundler` would also feature a (giant param list) ctor for direct use.

`org.trellisldp.app.SimpleServiceBundler` would subclass `BaseServiceBundler` as would `org.trellisldp.osgi.TrellisServiceBundler`, `org.trellisldp.app.triplestore.TrellisServiceBundler` and `edu.si.trellis.CassandraServiceBundler`. All would lose a great of boilerplate and instead cut down to focusing on the logic that is really unique to their contexts. Spring understands the same annotations, so a Spring app might be able to just use `BaseServiceBundler` directly, impling only `ConstraintServices`. (The fact that we need to have different implementations for `ConstraintServices` is because sadly, one place that Spring does better than CDI is by letting you inject a collection directly. Of course, CDI fans would say that those are loose semantics and for best long-term results, you should make your semantically-precise type anyway, which is what we would be doing by inventing `ConstraintServices`.

Do you see what I mean? Sorry for the long comment… I can try to offer a PR to this PR, but I'm not sure I can deal with the Gradle stuff. At least I can send you some pieces of this, if you want.

acoburn:
@ajs6f thanks for the feedback. I'm definitely not tied to this design; my main goal was with reusability, and your suggestion seems to have a greater level of reusability. I also like your suggestion of putting this in `org.trellisldp.app` as a base class.

Are you thinking that the `ConstraintServices` type would also go into the `o.t.app` module?

ajs6f:
> Are you thinking that the ConstraintServices type would also go into the o.t.app module?
 
Yup, exactly. The idea is to get to this; to build a new backend or new impls of core componentry like `EventService`, you need look only at `-api`. To build a new deployable, you need look only at `-app`. We could actually include instructions saying:

To build a new backend, impl `ResourceService` and maybe `BinaryService`. To build a new app, bring together Trellis and your backend using your build management of choice, then impl `ServiceBundler` (`extend BaseServiceBundler` to make it easy) and use `AHypotheticalTypeThatBringsTogetherNeededJAX-RSStuffTheWayServiceBundlerBringsTogetherNeededTrellisStuff`.

Maybe we can bikeshed the names a bit.

acoburn:
I've reworked the cdi-based `ServiceBundler` as an abstract base class. I haven't added the `ConstraintServices` interface part or a giant constructor, but this gives us a start.

components/app/src/main/java/org/trellisldp/app/BaseServiceBundler.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+package org.trellisldp.app;
+
+import javax.inject.Inject;
+
+import org.trellisldp.api.*;
+import org.trellisldp.http.core.*;
+
+/**
+ * A ServiceBundler that supplies injected components for a Trellis application.
```
ajs6f: How about 
```
A {@link ServiceBundler} that supplies components for a Trellis application from its fields.
<p>In this class, the fields are expected to be filled by injection, but in subclasses the fields may be filled by any appropriate means.
```

acoburn:
Should the class be `AbstractBaseServiceBundler`? or maybe `AbstractServiceBundler`? Just `BaseServiceBundler` seems incorrect, since it's an abstract class.

ajs6f:
It is for now, but if we shift to using a `ConstraintServices` type it could be entirely concrete, using a ctor. So my answer is whatevs, because ideally we'll move beyond that question soon.

ajs6f:
Sorry, I should have been more clear-- my answer is whatevs, with the expectation that we will end up with a _concrete_ `BaseServiceBundler`. At some point (not now!) I'm going to open the discussion about unbundling some of the components and letting things like `TrellisHttpResource` accept them directly in their ctors. (That goes along with the story of "shredding" `TrellisHttpResource` into per-HTTP-method classes, which I am going to do just as soon as I have time, so maybe in 2038).

acoburn:
@ajs6f this is ready for review.

Issue #465, opened 2019-08-09
Accept-Datetime should return the closest Memento


gregjan:
According to RFC 7089, section 4.5.3, the Accept-Datetime header can request a date from a TimeGate prior to the first memento or after the last memento. In these cases, the specification says that the closest Memento should be selected. (Relates to #188 )

acoburn:
Indeed -- in that case, the response should definitely be the first memento.

ajs6f:
Just a note that this just bit me in the butt (the change requires backend changes). It might be good to document this behavior requirement in the Javadocs for `MementoService::get`.

PR #468, opened 2019-08-09
Fix TimeGate/Memento handling

acoburn:
Resovles #465

For datetime negotiation requests that ask for a memento that predates
any existing memento, the Memento specification requires that the first
available Memento resource be returned.

https://tools.ietf.org/html/rfc7089#section-4.5.3

### DIFF
--- a/components/file/src/main/java/org/trellisldp/file/FileMementoService.java
+++ b/components/file/src/main/java/org/trellisldp/file/FileMementoService.java
@@ -100,9 +100,16 @@ public class FileMementoService implements MementoService {
             if (file.exists()) {
                 return new FileResource(identifier, file);
             }
-            final SortedSet<Instant> possible = listMementos(identifier).headSet(mementoTime);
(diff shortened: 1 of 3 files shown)


acoburn:
@gregjan this fixes the issue you brought up.

gregjan:
Okay, this looks right to me. I wanted to confirm that it will not fix the equivalent behavior on the trellis-ext-db version, correct?

acoburn:
Actually, this will fix the equivalent behavior in `trellis-ext-db`, though I will need to trigger a regeneration of the docker image

gregjan:
Oh, now I see. The DBMementoService ends up wrapping the FileMementoService.. I regenerated the trellis-ext-db docker image locally, after a trellis mvn install, and with the --refresh-dependencies flag. However, this Fedora test code is still getting a 406 on the time gate with a 1970s date time. I'll build it all again and double-check, now that it is merged to master.

PR #470, opened 2019-08-10
Microprofile deployment option

### LINKED ISSUE #209: MicroProfile packaging and deployment

acoburn:
Resolves #209

- [X] More JAX-RS filters need to be added to the `WebApplication` class
- [X] The liberty plugin is overly assiduous, cleaning things up afterwards, including any generated artifacts
- [x] The tests are extremely basic and need to be expanded
- [x] The `@Alternative` beans are not being automatically selected and therefore need a `@Produces` annotation, which seems incorrect, since they are listed as `<alternatives>` in the beans.xml file
- [x] We should use the `microprofile` feature instead of individual sub-features in the liberty config.
- [x] A `HealthCheck` implementation needs to be added to the `TriplestoreResourceServer`
- [x] Add memento tests


ajs6f:
Should I get rid of `trellis-microprofile` in favor of this?

platform/microprofile/src/test/liberty/config/server.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<server description="Trellis Server">
+    <featureManager>
+      <feature>cdi-2.0</feature>
```
ajs6f: Is there a reason you are using individual feature selections (as opposed to `microProfile-n.n`)?

acoburn: That's part of the WIP-ness of this branch. Here, I am using individual feature selections just to avoid the OpenAPI errors that are generated when using the full microProfile feature. Eventually, yes, we would want to just reference `microProfile-n.m`.

acoburn:
Let's keep your `trellis-microprofile` project around for the time being. It has been incredibly useful to me, and there are some DB-specific features that I will want to copy over to the `trellis-ext-db` project once we get this all sorted out.

platform/microprofile/src/test/liberty/config/server.xml
ajs6f: Cool.

ajs6f:
Ok, as usual I started work and went MIA, so I'll follow this and leave `t-m` be for now.

platform/microprofile/src/test/java/org/trellisldp/microprofile/SmokeIT.java
```diff
@@ -35,7 +35,7 @@
     private static final HttpClient testClient = HttpClients.createDefault();
 
     @Test
-    void findTrellis() throws IOException {
+    public void findTrellis() throws IOException {
```
ajs6f: Hm, why `public` here?

platform/microprofile/src/main/java/org/trellisldp/microprofile/TrellisServiceSupplier.java
```diff
@@ -32,19 +32,19 @@
  * triplestore resource service.
  */
 @ApplicationScoped
-class TrellisServiceSupplier {
+public class TrellisServiceSupplier {
 
     @Produces
-    private RDFConnection rdfConnection;
+    protected RDFConnection rdfConnection;
```
ajs6f: Why `protected` here? Do we expect subclasses of this?

acoburn: Now that you ask, I don't really have a good rationale for that -- maybe `private` would be better.

platform/microprofile/src/test/java/org/trellisldp/microprofile/SmokeIT.java
acoburn: For junit tests, I typically make the test methods `public`

platform/microprofile/src/main/java/org/trellisldp/microprofile/TrellisServiceSupplier.java
ajs6f: I'd go `private` unless/until there is a specific reason to do otherwise, but I'm old-fashioned like that.

acoburn: no problem, I'll change that back.

platform/microprofile/src/test/java/org/trellisldp/microprofile/SmokeIT.java
ajs6f: Yeah, why? JUnit doesn't need it (and on a minor note, some IDEs can warn on missing or malformed Javadocs, and making test methods `public` tends to cause a lot of false positives for that).

acoburn: Interesting. I'll revert that, too, though I'll be completely re-implementing the test code before this gets merged anyway.

ajs6f:
> The @Alternative beans are not being automatically selected and therefore need a @Produces annotation, which seems incorrect, since they are listed as <alternatives> in the beans.xml file

One thing to check-- the correct location of `beans.xml`. For a JAR archive, it should be in `META-INF`, but for a WAR archive, [it should be in `WEB-INF`](https://docs.oracle.com/javaee/6/tutorial/doc/gjbnz.html).

ajs6f:
I'm making up a commit now to move `beans.xml`.

acoburn:
Back to using open liberty.... I now have the trellis LDP tests wired in and passing (mostly). The two tests that fail relate to some sort of oddness in the CXF `ResponseBuilder` on a root resource such that http headers are overwritten rather than added. That is, adding a second `Link` header overwrites the first one instead of adding the second. This means that the check for the standard LDP headers on the root resource will fail, since those required headers are overwritten by other link headers added to the resource.

Injection is still only mostly working, since the `MementoService` and `EventService` need to be wired in manually. Still, this is good progress.

acoburn:
Note: in this PR, I am not using the `trellis-app` module, because it has dependencies on dropwizard, which in turn brings in jersey as a JAX-RS implementation. This interferes with open liberty, since liberty uses CXF. Once #472 is addressed, then we can start using the classes in `trellis-app`.

acoburn:
@ajs6f this is now ready for your review. I would still like to add a real notification service and add support for microprofile-jwt, but that can happen separately. This, at least, gives us the basic structure for µProfile in Trellis.

ajs6f:
Cool, still in Sweden for another day or so but will look at this on Monday
certainly.

platform/microprofile/src/main/java/org/trellisldp/microprofile/TrellisServiceSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.enterprise.inject.Produces;
+
+import org.apache.jena.rdfconnection.RDFConnection;
+import org.trellisldp.api.*;
+
+/**
+ * A managed bean that generates an RDF connection for the triplestore resource service.
+ */
+@ApplicationScoped
+public class TrellisServiceSupplier {
```
ajs6f: Does this belong in the main µprofile artifact, or in something specific to the triplestore backend? If I want to reuse the main artifact, am I going to have to work around this in order to use my `EventService` of choice?

platform/microprofile/src/main/webapp/WEB-INF/beans.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd" version="1.1"
+  bean-discovery-mode="all">
+  <alternatives>
+     <class>org.trellisldp.api.NoopEventService</class>
```
ajs6f: I'm not sure why we need this _and_ the `@Produces` method above?

platform/microprofile/src/main/java/org/trellisldp/microprofile/WebApplication.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public Set<Object> getSingletons() {
+        return asList(httpResource, httpFilter, corsFilter, cacheFilter, agentFilter, oauthFilter).stream()
+            .collect(toSet());
+    }
+
+    private void printBanner(final String name) {
+        final URL resource = Thread.currentThread().getContextClassLoader().getResource("banner.txt");
```
ajs6f: This seems a bit fragile to me. Can we not inline the banner in this class? If not, just how big is it?

acoburn: I have the `banner.txt` resource in the `trellis-app` artifact, which gets used by both the dropwizard and µprofile artifacts. The code here is pretty much just a simplification of how DropWizard prints out the banner.

platform/microprofile/src/main/webapp/WEB-INF/beans.xml
acoburn: You are correct, this is not needed.

platform/microprofile/src/main/java/org/trellisldp/microprofile/WebApplication.java
ajs6f: I'm not going to lie down in the road over this. I've just seen too many troubles come from loading resources from the "current thread" classpath to feel super comfortable, so I'd rather see it inlined, but until we actually have a problem, whatevs.

platform/microprofile/src/main/java/org/trellisldp/microprofile/TrellisServiceSupplier.java
acoburn: This class is definitely specific to the triplestore back-end with a particular selection of service implementations. What I was trying to achieve here was a specific microprofile-based deployment that could be used as an example for other µP applications. The `trellis-app` is really intended to be the shared layer that is built upon for µP, dropwizard, spring, etc.

ajs6f: Ok, that's a good strategy, but in that case, I think we should be more upfront about this artifact being an _example_ and not a _reusable part_. Maybe a name change `microprofile`->`triplestore`? Given that we expect to deprecate Dropwizard eventually, do we even need to call out µmicroprofile explicitly?

acoburn: That's a good point. I'd want to be able to distinguish between this and the `./components/triplestore` project (and also the `trellis-triplestore` artifact). With the Dropwizard-based code, I've used `app-triplestore` and similar as a naming convention. For now, how about `mp-triplestore` or `app-mp-triplestore`? Or maybe some other name?

platform/microprofile/README.md
```diff
@@ -0,0 +1,7 @@
+# Trellis Deployable Application
```
ajs6f: Ok, I'm being pushy here, but can we explicitly mention that this artifact uses the triplestore backend? I'm thinking of the casual browser who is just trying to figure out what different modules do.

platform/microprofile/src/main/java/org/trellisldp/microprofile/WebApplication.java
acoburn: I've updated this slightly to (hopefully) make this less fragile.

platform/microprofile/README.md
acoburn: Yes, I'll fill that out some more.

platform/microprofile/src/main/java/org/trellisldp/microprofile/TrellisServiceSupplier.java
ajs6f: I like `app-mp-triplestore` because it has all the important facts. Let's roll with that!

acoburn:
I added a few minor touch-ups to the code, adjusting the gradle build as per the suggestion of @jholleran and adding a `@PostConstruct` annotation to the `TriplestoreResourceService::initialize` method.

I'll let this sit for a few more days in case there are other comments; otherwise, I'll plan to merge this early next week.

PR #471, opened 2019-08-10
Put beans.xml into WEB-INF

ajs6f:
https://github.com/trellis-ldp/trellis/pull/470


ajs6f:
I left the `RDFConnection` as a producer field because of the config required, but we might be able to move it out later. Not an immediate problem, I reckon.

acoburn:
I tried this exact thing earlier, and the problem is that the application doesn't start. The log in `./build/wlp/usr/servers/trellis/logs/ffdc/ffdc*.log` shows this error:

```
org.jboss.weld.exceptions.DeploymentException: WELD-001408: Unsatisfied dependencies for type MementoService with qualifiers @Default
```

I also tried putting the `beans.xml` file into `./usr/main/webapp/WEB-INF`, but no dice there, either. In this latter case, the exploded WAR has the `beans.xml` file located at `./WEB-INF/beans.xml`

ajs6f:
Hm. Then we have a different problem, because my build completes (with tests) with this change, so the test isn't actually testing what it should. But I don't see how that would be occurring. Can you give me a more complete log? `WEB-INF/beans.xml` is without doubt the correct place, and if the `alternative` element in there referenced a non-existent or otherwise problematic class, it should die on wiring validation. Additionally, I would think the `EventService` injection would also have a problem, but that could be being hidden by the other gap.

ajs6f:
Notice also the CI builds appear to complete… so am I misunderstanding what you say?

acoburn:
The tests are really not testing anything at the moment -- they check that the container comes up, but that's it. You can manually look at the test output to see if the response headers are LDP (i.e. Trellis) headers or whether they are just Open Liberty webroot headers. (I know that I still need to add real tests to this)

ajs6f:
Okay, but according to your error above, the container _shouldn't_ be coming up. A missing injection slot should shut the container down. I'll try and reproduce that somehow, but perhaps you can tell me how you got the error? (I.e. using a as-yet-unseen test or some other procedure?)

acoburn:
Yes, the `EventService` would also have a similar issue here, but the deployment stage fails on the first issue it encounters, so we don't see that. You can see the full log here: https://gist.github.com/acoburn/d9d51bf7045aae26be55d53db4afa5c8

The web application (trellis) doesn't come up, but Open Liberty, itself, does, and it is available on port 9080.

I should probably add in some real tests before iterating on this.

acoburn:
As for how I get that test error:

```
$ ./gradlew check
```

Then, look in this directory: `./build/wlp/usr/servers/trellis/logs/`, especially the `logs/ffdc` directory.

ajs6f:
Yup, that's what it looks like. I think we have to fix this before going ahead-- manually wiring via producer fields kind of negates the whole point of injection. I'll try to duplicate and investigate.

ajs6f:
Ok, take a look (after a build) in : `platform/microprofile/build/wlp/usr/servers/trellis/apps/trellis-microprofile.war.xml` and notice the line
```
````

acoburn:
I have added tests from the `trellis-test` module, and I am finding that OpenLiberty seems to modify the response headers such that only the final `Link` header is produced. I don't know if there is some setting that needs to be adjusted, but I am starting to think about using a different backend implementation (quarkus, tomEE, etc)

ajs6f:
Ok by me. You know how far I got with OL in `trellis-micro` and nothing I saw made me determined to commit to it. Payara also looked attractive. Be aware that Quarkus (in order to be able to produce native executables) doesn't implement the full CDI spec. We would run into [some of its limitations](https://quarkus.io/guides/cdi-reference#limitations).

ajs6f:
Also note that Quarkus asks us for some weird things, like [to avoid `private` members](https://quarkus.io/guides/cdi-reference)!

acoburn:
I'm going to close this because I've effectively added this exact modification to the microprofile PR (and it is now working!)

PR #487, opened 2019-08-28
Add ApplicationScoped annotations

### LINKED ISSUE #486: Additional CDI annotations
In working with Quarkus, I am finding that injection only works for beans that are annotated with `@ApplicationScoped`, so we should add that annotation to all the application-scoped beans.

acoburn:
Resolves #486


components/audit/src/main/java/org/trellisldp/audit/DefaultAuditService.java
```diff
@@ -41,6 +43,7 @@
  * 
  * @author acoburn
  */
+@ApplicationScoped
```
jholleran: Can/Should @ApplicationScoped be on the NoopAuditService also? And possibly all the Noop*Service's.

PR #491, opened 2019-08-30
Remove final keyword to no-op services

acoburn:
Quarkus doesn't like `@ApplicationScoped` beans that have the `final` keyword. This just removes those keywords.


acoburn:
No, it just fails rather unglamorously ;-)

Issue #495, opened 2019-08-31
Review JAX-RS injection mechanisms


acoburn:
The JAX-RS classes (resource matches and filters) currently make use of constructor injection, which makes unit testing really straight-forward, and this works via CDI injection as expected with both Jersey and CXF. RESTeasy, however, does not support constructor injection, which makes it harder to use the `TrellisHttpResource` class out-of-the-box with Quarkus.

As an alternative, we could change the `TrellisHttpResource` and `AgentAuthorizationFilter` classes to support either field injection or setter injection. (I would opt for setter injection.)

The other services at the Trellis API level can continue to use ctor injection without issue.

acoburn:
I was thinking about this a bit today, and if we move toward either field or setter injection, we could possibly dispense with the `ServiceBundler` construct. The only purpose of that interface was to avoid having constructors with dozens of arguments. It would also become easier to break the `TrellisHttpResource` class into smaller units (which @ajs6f has suggested in the past)

ajs6f:
We can remove `ServiceBundler` by factoring, as you rightly say, so I don't think it makes an argument to introduce a new kind of injection.

PR #500, opened 2019-09-02
Adjust injection mechanism for JAX-RS beans

acoburn:
Resovles #495

All non-JAX-RS beans continue to use constructor injection

### DIFF
--- a/components/webac/src/main/java/org/trellisldp/webac/WebAcFilter.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/WebAcFilter.java
@@ -108,20 +108,30 @@ public class WebAcFilter implements ContainerRequestFilter, ContainerResponseFil
     private static final Set<String> writable = new HashSet<>(asList("PUT", "PATCH", "DELETE"));
     private static final Set<String> appendable = new HashSet<>(asList("POST"));
 
+    protected final WebAcService accessService;
     private final List<String> challenges;
     private final String baseUrl;
 
-    protected WebAcService accessService;
+    /**
+     * No-op constructor for RESTeasy and CDI.
+     */
+    public WebAcFilter() {
+        this(null);
+    }
 
     /**
      * Create a new WebAc-based auth filter.
+     *
+     * @param accessService the access service
      */
-    public WebAcFilter() {
-        this(getConfig());
+    @Inject
+    public WebAcFilter(final WebAcService accessService) {
+        this(accessService, getConfig());
     }
 
-    private WebAcFilter(final Config config) {
-        this(asList(config.getOptionalValue(CONFIG_WEBAC_CHALLENGES, String.class).orElse("").split(",")),
+    private WebAcFilter(final WebAcService accessService, final Config config) {
+        this(accessService,
+                asList(config.getOptionalValue(CONFIG_WEBAC_CHALLENGES, String.class).orElse("").split(",")),
                 config.getOptionalValue(CONFIG_WEBAC_REALM, String.class).orElse("trellis"),
                 config.getOptionalValue(CONFIG_HTTP_BASE_URL, String.class).orElse(null));
     }
@@ -129,13 +139,16 @@ public class WebAcFilter implements ContainerRequestFilter, ContainerResponseFil
     /**
      * Create a WebAc-based auth filter.
      *
+     * @param accessService the access service
      * @param challengeTypes the WWW-Authenticate challenge types
      * @param realm the authentication realm
      * @param baseUrl the base URL, may be null
      */
-    public WebAcFilter(final List<String> challengeTypes, final String realm, final String baseUrl) {
+    public WebAcFilter(final WebAcService accessService, final List<String> challengeTypes, final String realm,
+            final String baseUrl) {
         requireNonNull(challengeTypes, "Challenges may not be null!");
         requireNonNull(realm, "Realm may not be null!");
+        this.accessService = accessService;
         this.challenges = challengeTypes.stream().map(String::trim).map(ch -> ch + " realm=\"" + realm + "\"")
             .collect(toList());
         this.baseUrl = baseUrl;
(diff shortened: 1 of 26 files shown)


components/webac/src/main/java/org/trellisldp/webac/WebAcFilter.java
```diff
@@ -163,6 +148,16 @@ public WebAcFilter(final WebAcService accessService, final List<String> challeng
                 stream(a.split(",")).map(String::trim).map(String::toUpperCase).forEach(appendable::add));
     }
 
+    /**
+     * Set the access service.
+     *
+     * @param accessService the access service
+     */
+    @Inject
+    public void setAccessService(final WebAcService accessService) {
```
jholleran: If someone wants to use these classes in a non CDI env changing this to a setter makes this class harder to use. There will be a NPE in the filter if they forget to set this dependency. Do you see any issue here? Should there be a default or Noop implementation of this dependency set in the constructor?

acoburn: This is exactly why I prefer ctor injection. The issue, however, is that the JAX-RS specification (independent of CDI) does not require support for ctor injection. So while Jersey and CXF both support ctor injection, RESTeasy does not, which actually makes these classes rather difficult to use with Thorntail, Quarkus or anything else built on RESTeasy.

So, in order to support the widest range of JAX-RS implementations, we will need either field or setter injection for these classes, and I find setter injection to be much easier to deal with than field injection.

That said, I don't have an opinion about adding another constructor that also sets this field. I would be a little concerned that it would be a bit confusing: do you set the object in the ctor or don't you? Certainly the documentation could clarify what is going on.

Is it reasonable to expect that someone using JAX-RS will be familiar with how injection works?

acoburn:
@gregjan this PR should make your work with Thorntail considerably easier

gregjan:
Thanks! I will have a look next hour.
G

acoburn:
@gregjan I think we _have_ encountered an issue with RESTeasy in that we've had to create no-arg ctors such as https://github.com/trellis-ldp/trellis/blob/master/core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java#L121-L126 in order to make injection work properly.

This change means that those no-arg ctors are actually usable (provided that the setter injection is handled by the DI framework or manually)

gregjan:
Ahaa. Now I get it and it makes sense. I'm guessing that this is orthogonal to the JMSEventService CDI stuff? I going to create a `@Produces` method for a JMS connection, which would go in the JMS component.

acoburn:
@ajs6f the short answer is that with this change I am able to get Trellis to work with Quarkus; otherwise, I am entirely unable to get it to work.

ajs6f:
That's a pretty short answer. 😁 Can I ask whether you are working against a Quarkus that includes the fixes they've done for this issue?

acoburn:
The fixes for this issue that Quarkus put in place have been available since 0.14. I am using 0.21.1, which is the current release.

I think part of the issue is that, in order to add support for Thorntail, we added some non-functional no-arg constructors to various JAX-RS beans, yet those no-arg ctors set a lot of the internal fields to `null`. Whether that is the source of the issues, however, is a little hard to ascertain: the stacktraces are entirely unuseful.

ajs6f:
Hm. Quarkus is intruding on our code (e.g. w/o ctor injection it's rather more difficult to make fields `final`) in ways that other toolkits haven't. In this PR we  have whole sections of code that exist only to support Quarkus. This is why I didn't choose it to work with (this really means we should stop calling it "microprofile app" and call it "Quarkus app" instead, although I'm not seriously suggesting that), although the promise of more ergonomic build machinery is certainly attractive.

Can we hold off on this for two weeks? I understand that's a long time and I don't know what Inrupt has going on that might make that easier or harder, but I'm traveling all next week (ApacheCon) and I'd be happy to dig into this when I get back. Debugging CDI is indeed a real pain in the neck sometimes, especially with a newish toolkit like Quarkus. But I'd really like to get this right, and by all advertisement of the projects involved, it should work. If it really doesn't, we do what we have to do, but we can also then file something back upstream, which would be neighborly.

acoburn:
We can definitely hold off on this for a bit. That will also give me more time to figure out what is going on.

ajs6f:
Thanks @acoburn! I'm really not trying to be obstructive here-- I really do think we should try to stay with ctor injection if at all possible because it makes for much more readable code over the long term, IMO.

acoburn:
I would also strongly prefer to use ctor injection. I will see if I can get this all to work with that structure. Stay tuned.

acoburn:
I have pushed two new commits to this branch. The first reverts the ctor injection changes, so we're back to ctor-injection all the time (which I now have working with Quarkus). There are some small changes needed, as you will see, particularly with making the no-arg ctors `public`.

The last commit here changes the injection point for the WebAC service such that it accepts a `ServiceBundler` (in addition to the existing `ResourceService`).

Otherwise, this is now ready for re-review

acoburn:
Yes, the ctors had to be public. I tried setting them to `protected` (as well as using the existing `package-protected` visibility level), but that just resulted in a big stack trace:

```
Caused by: java.lang.RuntimeException: java.lang.RuntimeException: RESTEASY003190: Could not find constructor for class: org.trellisldp.http.TrellisHttpResource
```

ajs6f:
I may look into that, because I'm curious about Quarkus' limits. They _show_ package-private ctors [in use in examples](https://quarkus.io/guides/cdi-reference#private-members) (second example, after "Or constructor injection:"), but what does that prove? 

To be clear, I am totally cool with this PR, just curious about this new tool we are working with.

acoburn:
From what I've seen, the package-protected fields/methods/ctors are all fine inside the package in which Quarkus is operating, but for packages in external Jars, that's where the `public` visibility needs to be in place. Or at least so it seems.

PR #502, opened 2019-09-04
Add support for BOM generation

### LINKED ISSUE #501: Create a BOM artifact
A maven BOM artifact could make it nicer to work with trellis components

acoburn:
Resolves #501

There's no java code here, just gradle.


acoburn:
I tested this locally with the `trellis-ext-db` project, and it seems to work correctly.

PR #503, opened 2019-09-04
[WIP] Add revision tag to Resource and Metadata classes

### LINKED ISSUE #497: Extend Metadata class
The metadata class currently has not slot for a "revision" value, which could be really useful for backends that use this for concurrency control. Effectively, this could be either an `Optional` or nullable value that is generated using the `EtagGenerator` interface. Backends would not necessarily need to use this (the Triplestore and DB backends would ignore the value), but I can imagine other backends that could require this.

acoburn:
*Please do not merge*

Resolves #497

This adds a revision method to the `Metadata` class and the `Resource` interface. It does not do anything with these values in the HTTP layer (they will eventually be connected to the handling of ETags)

Some open questions:
1. Should `Resource::getRevision` return `Optional<String>` or do we expect a value there (I think we expect a value)
2. Should the `Metadata::getRevision` return `Optional<String>`? I notice an inconsistency in the API (it seems the two methods should align)
3. With this change, can we entirely remove the `org.trellisldp.http.core.ETagGenerator` interface? (I think yes, especially if we are not returning `Optional` values)


components/file/src/main/java/org/trellisldp/file/FileResource.java
```diff
@@ -78,6 +79,11 @@ public Instant getModified() {
         return asLiteral(DC.modified).map(Instant::parse).orElse(null);
     }
 
+    @Override
+    public String getRevision() {
+        return md5Hex(getModified().getNano() + "." + identifier);
```
ajs6f: Why hash this? It's not getting any shorter. If two `getModified().getNano() + "." + identifier`s collide, so will their hash, so we're not getting anything other than opacity…

components/test/src/main/java/org/trellisldp/test/InMemoryResourceService.java
```diff
@@ -196,6 +197,11 @@ public Instant getModified() {
             return modified;
         }
 
+        @Override
+        public String getRevision() {
+            return md5Hex(modified.getNano() + "." + identifier);
```
ajs6f: Ditto about hashing.

core/api/src/main/java/org/trellisldp/api/Metadata.java
```diff
@@ -45,11 +46,12 @@
      * @param memberRelation an LDP hasMemberRelation predicate, may be {@code null}
      * @param memberOfRelation an LDP isMemberOfRelation predicate, may be {@code null}
      * @param insertedContentRelation an LDP insertedContentRelation, may be {@code null}
+     * @param revision a revision value, may be {@code null}
```
ajs6f: How about a Javadoc link to the doc for `Resource::getRevision`? I think that would help people reading the code.

core/api/src/main/java/org/trellisldp/api/Resource.java
```diff
@@ -133,6 +143,18 @@ public String toString() {
...
+     * Get the revision or state tag of the resource.
+     *
+     * @apiNote the return value of this method should be unique to the state of the resource.
+     *          That is, it should be unique across all resources on the server, and whenever
+     *          a given resource changes, this revision tag should also change. To create this
+     *          value, an implementation may simply generate a hash of the identifier and
+     *          modification date.
+     * @return the revision tag of the resource
+     */
+    String getRevision();
```
ajs6f: Hm. Better `Optional<String>` here? Also, the hash thing again. If the purpose of this member is solely to _differentiate_ (not, for example, to abbreviate _and_ differentiate, or to distribute), I don't see the value of hashing.

core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
```diff
@@ -191,13 +191,13 @@ public void initialize() {
         if (MISSING_RESOURCE.equals(res) || DELETED_RESOURCE.equals(res)) {
             LOGGER.info("Initializing root container: {}", id);
             return trellis.getResourceService().create(Metadata.builder(id).interactionModel(LDP.BasicContainer)
-                    .build(), dataset);
+                    .hasAcl(true).build(), dataset);
```
ajs6f: Why `hasAcl(true)` suddenly?

components/file/src/main/java/org/trellisldp/file/FileResource.java
acoburn: Hashing is both for opacity and for conformance. The [spec](https://tools.ietf.org/html/rfc7232#section-2.3) explicitly calls out opacity and limits the valid characters to be ASCII. This way, if someone creates a resource with a unicode character (perfectly valid), it won't wreck the ETag handler.

core/api/src/main/java/org/trellisldp/api/Metadata.java
acoburn: Good idea.

components/file/src/main/java/org/trellisldp/file/FileResource.java
ajs6f: Whoa, wait! I did not understand this new method to be required to produce legal ETags and nothing in the Javadocs in this PR says so. Is that the requirement you are creating here?

core/api/src/main/java/org/trellisldp/api/Resource.java
acoburn: The problem with `Optional<String>` would be that the HTTP layer would then need to handle the ETag generation. We could re-add the `EtagGenerator` interface for cases where the `Resource` doesn't respond with a revision value. Note: at present, I'm assuming that this response will always be non-null

ajs6f: Ok, but that is not what you have documented here at all. This isn't `getRevision`, it's `getETag`. That is a different move entirely, and if you want to do that we better document it. Nothing in this PR gave me to understand that this method is expected to produce an RFC7232 ETag.

core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
acoburn: That was an oversight. I could add that to a separate commit, if you prefer. But this initialization step explicitly adds an ACL resource to this root resource, so it will always have an ACL present. Depending on how the persistence backend handles this value, it is more correct to explicitly state `hasAcl(true)` in the builder

components/file/src/main/java/org/trellisldp/file/FileResource.java
acoburn: That would be part of what's going on. But you raise an important point -- we could easily move the hashing part into the HTTP layer and set a `default` method for `Resource::getRevision` simply as a concatenation of `Resource::getModified` and `Resource::getIdentifier`

core/http/src/main/java/org/trellisldp/http/TrellisHttpResource.java
ajs6f: It's cool, just wasn't sure why it was here, but it's here for good reason.

core/api/src/main/java/org/trellisldp/api/Resource.java
acoburn: I'm going to re-work this a bit so that `::getRevision` is not expected to return a conforming ETag value

components/file/src/main/java/org/trellisldp/file/FileResource.java
ajs6f: Absolutely (and on my current understanding, preferable). Here are some options:
1. Keep an injectable ETag generator that is smarter.
    * Pro: keeps all the HTTP-standard-specific  stuff in one place, `trellis-http`, con: that generator has to get smarter and smarter for every new use case.
2. What you have here (while perhaps renaming `getRevision` to `getETag` or at least documenting it more explicitly)
    * Pro: keeps the backend-specific smarts in the backend where they belong, con: forces everyone to repeat at least some logic for ETags.
3. What you just suggested (let `Resource`s return the raw material for ETags and let the HTTP layer combine/transform it to ensure legal syntax)
    * Pro: separates responsibilities nicely, keeps things DRY, con: not much that I can see.

core/api/src/main/java/org/trellisldp/api/Resource.java
ajs6f: Right, see https://github.com/trellis-ldp/trellis/pull/503#discussion_r321347455 and my reply. I think you're making the right move.

PR #505, opened 2019-09-06
Moves from ACL database initialization to a permissive default root ACL

gregjan:
Takes out database initialization of the root ACL, leaving WebAC considerations outside of persistence backend initialization. Instead there is a default ACL that permits all actions by all agents, i.e. no restrictions. This is a single Authorization that is held in a static variable.

### DIFF
--- a/components/webac/src/main/java/org/trellisldp/webac/WebAcService.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/WebAcService.java
@@ -23,6 +23,7 @@ import static java.util.Collections.singletonList;
 import static java.util.Collections.unmodifiableList;
 import static java.util.Collections.unmodifiableSet;
 import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.CompletableFuture.runAsync;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.eclipse.microprofile.config.ConfigProvider.getConfig;
@@ -44,9 +45,11 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
+import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
+import org.apache.commons.rdf.api.Dataset;
 import org.apache.commons.rdf.api.Graph;
 import org.apache.commons.rdf.api.IRI;
 import org.apache.commons.rdf.api.Quad;
@@ -55,6 +58,7 @@ import org.apache.commons.rdf.api.RDFTerm;
 import org.apache.commons.rdf.api.Triple;
 import org.slf4j.Logger;
 import org.trellisldp.api.CacheService;
+import org.trellisldp.api.Metadata;
 import org.trellisldp.api.NoopResourceService;
 import org.trellisldp.api.Resource;
 import org.trellisldp.api.ResourceService;
@@ -88,7 +92,8 @@ public class WebAcService {
 
     /** The permissive Authorizations in effect when no ACL is present on the root node. **/
     private static final List<Authorization> defaultRootAuthorizations =
-        unmodifiableList(generateDefaultRootAuthorizations());
+        unmodifiableList(singletonList(Authorization.from(rootAuth, generateDefaultRootAuthorizations()
+            .getGraph(Trellis.PreferAccessControl).get())));
 
     static {
         allModes.add(ACL.Read);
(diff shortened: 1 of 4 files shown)


acoburn:
I really like where this is heading: it moves the ACL initialization logic outside of the HTTP layer and isolates it in the WebAC layer. It also provides a "backstop" in case the root resource ACL is not present. That said, we lose any materialization of the root ACL resource itself, which is valuable for specification conformance.

Here's a specific suggestion: rework the static method in the `WebAcService` class, first making that method `public` (for better re-use elsewhere, if needed), and second have it return a `Graph`. Then, a second static method can convert that `Graph` into an `Authorization` object, which can be used to populate the `defaultRootAuthorizations` field. Separately, create a method:

```java
@PostConstruct
public void initialize() {
    // fetch the root resource, check `Resource::hasAcl`
    // if it is empty, populate that resource as is done currently in the HTTP layer,
    // using the `Graph` (or `Dataset`, since it might be more convenient to use datasets everywhere) created above (plus any existing user-managed triples)
}
```

Then, in the `dropwizard` [component](https://github.com/trellis-ldp/trellis/blob/master/components/dropwizard/src/main/java/org/trellisldp/dropwizard/AbstractTrellisApplication.java#L138), be sure to call `WebAcService::initialize`.

Finally, there are some test methods in the HTTP layer that will need adjustment because they are failing with this change.

gregjan:
Made changes as you suggested, using a Dataset to represent the root ACL. Tests pass now, as the TrellisHttpResource is no longer responsible for the root ACL provisioning. Dropwizard should initialize the root node first, then add the ACL.

components/webac/src/main/java/org/trellisldp/webac/WebAcService.java
```diff
@@ -108,18 +113,44 @@ public WebAcService() {
...
+
+    private CompletionStage<Void> initialize(final IRI id, final Resource res, final Dataset dataset) {
+        if (!res.hasAcl()) {
+            LOGGER.info("Initializing root ACL: {}", id);
+            try (final Stream<Quad> quads = res.stream(Trellis.PreferUserManaged)) {
+                quads.forEach(dataset::add);
+            }
+            return this.resourceService.replace(Metadata.builder(res).build(), dataset);
+        } else {
+            return runAsync(() -> { });
```
ajs6f: I would return a `CompletableFuture.completedFuture(null)`. Also I would take that out as a constant called `DONE` or the like.

ajs6f: Also I would log (at `INFO`) something here, like "No changes made to pre-existing root ACL."

gregjan: I was wondering what the shorthand should be there.. thanks.

components/webac/src/main/java/org/trellisldp/webac/WebAcService.java
```diff
@@ -79,10 +84,17 @@
...
     private static final Logger LOGGER = getLogger(WebAcService.class);
+    private static final CompletionStage<Void> DONE = CompletableFuture.completedFuture(null);
     private static final RDF rdf = getInstance();
     private static final IRI root = rdf.createIRI(TRELLIS_DATA_PREFIX);
+    private static final IRI rootAuth = rdf.createIRI(TRELLIS_DATA_PREFIX + "#auth");
     private static final Set<IRI> allModes = new HashSet<>();
 
+    /** The permissive Authorizations in effect when no ACL is present on the root node. **/
+    private static final List<Authorization> defaultRootAuthorizations =
+        unmodifiableList(singletonList(Authorization.from(rootAuth, generateDefaultRootAuthorizations()
```
acoburn: Minor: the `generateDefaultRootAuthorizations` method call here generates a `Dataset` but never closes it. Perhaps it would be better to create an additional `static` method such as:

```java
private static final List<Authorization> getAuthorizations() {
    try (final Dataset dataset = generateDefaultRootAuthorizations()) {
        dataset.getGraph(Trellis.PreferAccessControl).map(graph -> Authorization.from(rootAuth, graph)
                .map(Collections::singletonList).orElseGet(emptyList());
    } catch (final Exception ex) {
        throw new RuntimeTrellisException("...", ex);
    }
}
```

And then you can just call:

```java
private static final List<Authorization> defaultRootAuthorizations = unmodifiableList(getAuthorizations());
```

Feel free to change names, etc.

gregjan: I am reusing the Dataset for the root node ACL persistence and the "fall back ACL". So I didn't want to close it, because I keep it around. If you close it I believe you can no longer call getGraph(), etc...

gregjan: Aha, I think you are suggesting that I keep the authorizations and not the Dataset, which works for the fall back case just fine. Let me try that.

ajs6f: This is (really!) a good motivation for us to tackle https://github.com/trellis-ldp/trellis/issues/358 and https://github.com/orgs/trellis-ldp/projects/4. The truth is that `Dataset::close` is totally unclear. Its comments include ` The behaviour of the other dataset methods are undefined after closing the dataset.` Having said that, I've done nothing to act on this stuff, so I'll shut up now and go back and keep working on `trellis-cassandra`.

PR #507, opened 2019-09-07
Implement microprofile reactive messaging

### LINKED ISSUE #483: Implement an EventService using MP reactive messaging
Microprofile provides a generic message processing system with [reactive streams](https://github.com/eclipse/microprofile-reactive-messaging/blob/master/spec/src/main/asciidoc/architecture.asciidoc). There is a more complete description on the [SmallRye website](https://smallrye.io/smallrye-reactive-messaging/).

This would likely be an easy way to integrate an `EventService` component into a microprofile-based application while also being generic across message brokers.

acoburn:
Resolves #483

This is still a bit experimental, but it passes the tests and there is so little code here, that I am inclined to move this forward.

### DIFF
--- a/notifications/reactive/src/main/java/org/trellisldp/reactive/ReactiveEventService.java
+++ b/notifications/reactive/src/main/java/org/trellisldp/reactive/ReactiveEventService.java
@@ -36,7 +36,7 @@ public class ReactiveEventService implements EventService {
     private final PublishSubject<Message<String>> subject = PublishSubject.create();
 
     /**
-     * Create a new Kafka Event Service.
+     * Create a new Reactive Stream Event Service.
      * @param serializer the event serializer
      */
     @Inject
--- a/notifications/reactive/src/main/java/org/trellisldp/reactive/ReactiveEventService.java
+++ b/notifications/reactive/src/main/java/org/trellisldp/reactive/ReactiveEventService.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


notifications/reactive/src/test/java/org/trellisldp/reactive/ReactiveEventServiceTest.java
```diff
@@ -0,0 +1,117 @@
...
+        service.emit(mockEvent);
+        await().atMost(FIVE_SECONDS).until(() -> collector.getResults().size() == 1);
+        service.emit(mockEvent);
+        await().atMost(FIVE_SECONDS).until(() -> collector.getResults().size() == 2);
+        service.emit(mockEvent);
+        service.emit(mockEvent);
+        await().atMost(FIVE_SECONDS).until(() -> collector.getResults().size() == 4);
+    }
+
+    private <T> T getInstance(final Class<T> type) {
```
ajs6f: If the idea is to get properly set-up beans from the container, can we not just inject them? I'm not at all familiar with Weld's test gear so maybe not?

acoburn: I'm also not super familiar with Weld's test gear, but I'll see if there is something easy that will do that injection for us.

ajs6f: https://github.com/weld/weld-junit/tree/master/junit5

acoburn: That last commit should address this. It also makes the test code much more compact.

notifications/reactive/src/main/java/org/trellisldp/reactive/ReactiveEventService.java
```diff
@@ -0,0 +1,60 @@
...
+@ApplicationScoped
+public class ReactiveEventService implements EventService {
+
+    public static final String REACTIVE_DESTINATION = "trellis";
+
+    private final ActivityStreamService serializer;
+    private final PublishSubject<Message<String>> subject = PublishSubject.create();
+
+    /**
+     * Create a new Kafka Event Service.
```
ajs6f: Is this "Kafka" left over from some cut-and-paste?

acoburn: that's now fixed.

PR #510, opened 2019-09-11
Less asynchrony for Binary::getContent

ajs6f:
This changeset removes one level of asynchrony from the path between binary data and HTTP by switching the signatures of `Binary::getContent` from `CompletionStage<InputStream>` to `InputStream`. This is appropriate because the only use `trellis-http` makes use of the `CompletionStage<InputStream>` is to immediately call `toCompletableFuture().join()` on it, which is absurd. We should rely going forward on the practice we've developed of answering HTTP requests using `@Suspended` by getting a `CompletionStage` from one or more of `ResourceService`, `MementoService`, `BinaryService`, etc.. The types _inside_ those `CompletionStage`s would themselves feature methods that return ordinary (non-promise/future) types. This changeset causes that to be the case.

### DIFF
--- a/core/http/src/main/java/org/trellisldp/http/impl/GetHandler.java
+++ b/core/http/src/main/java/org/trellisldp/http/impl/GetHandler.java
@@ -16,6 +16,7 @@ package org.trellisldp.http.impl;
 import static java.lang.String.join;
 import static java.util.Collections.singletonList;
 import static java.util.Date.from;
+import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
@@ -200,7 +201,7 @@ public class GetHandler extends BaseLdpHandler {
      * @param builder the response builder
      * @return the response builder
      */
-    public ResponseBuilder getRepresentation(final ResponseBuilder builder) {
+    public CompletionStage<ResponseBuilder> getRepresentation(final ResponseBuilder builder) {
         // Add NonRDFSource-related "describe*" link headers, provided this isn't an ACL resource
         getResource().getBinaryMetadata().filter(ds -> !isAclRequest()).ifPresent(ds -> {
             final String base = getBaseBinaryIdentifier();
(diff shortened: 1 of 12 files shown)


core/http/src/main/java/org/trellisldp/http/impl/GetHandler.java
```diff
@@ -355,29 +356,22 @@ private ResponseBuilder getLdpNr(final ResponseBuilder builder) {
...
-                try (final InputStream binary = getBinaryStream(dsid, getRequest()).toCompletableFuture().join()) {
-                    IOUtils.copy(binary, out);
-                }
-            }
-        };
-
-        return builder.entity(stream);
+        // TODO -- with JDK 9 use InputStream::transferTo instead of IOUtils::copy
+        return getBinaryStream(dsid, getRequest())
+                        .thenApply(in -> (StreamingOutput) out -> IOUtils.copy(in, out))
```
acoburn: Should the input stream be closed here?

core/http/src/main/java/org/trellisldp/http/impl/GetHandler.java
```diff
@@ -355,29 +356,22 @@ private ResponseBuilder getLdpNr(final ResponseBuilder builder) {
...
-            public void write(final OutputStream out) throws IOException {
-                // TODO -- with JDK 9 use InputStream::transferTo instead of IOUtils::copy
-                try (final InputStream binary = getBinaryStream(dsid, getRequest()).toCompletableFuture().join()) {
-                    IOUtils.copy(binary, out);
-                }
-            }
-        };
-
-        return builder.entity(stream);
+        // TODO -- with JDK 9 use InputStream::transferTo instead of IOUtils::copy
```
acoburn: This TODO item can be removed. It is just cluttering things up, and I've been meaning to take it out. (Though feel free to leave it in, if you prefer)

ajs6f: Do you think this is invalid? I thought it was still useful…

acoburn: Sure, we can keep it here as a reminder

core/http/src/main/java/org/trellisldp/http/impl/GetHandler.java
ajs6f: Hm, I think you might be right. I'll take a look.

ajs6f: Good catch. I've fixed it.

PR #511, opened 2019-09-12
Add a delegatedBy identifier to admin sessions

acoburn:
When a user assumes the role of a super admin, this change adds the user's own webid to the `delegatedBy` property of the session. This means that the audit log will contain some additional (very useful) information.


acoburn:
Closing this, as it will be subsumed by #513

PR #513, opened 2019-09-14
Remove AgentService

### LINKED ISSUE #512: Consider removing AgentService
The `AgentService` interface does one thing: it converts the String from `java.security.Principal::getName` into an IRI. As it is, the entire code, effectively, consists of calling `RDF::createIRI`.

The advantage of the indirection is that one could, conceivably, inject a custom `AgentService` that performs an additional lookup or which follows certain business logic. (In practice, there appear to be no use cases for this level of indirection)

The counter-argument is that that sort of mapping should occur in the Authentication layer; having it elsewhere in the code base is confusing. As it is, we could easily place all the `AgentService` features directly in the HTTP (and WebDAV) layers. At the same time, the `trellis-agent` module could be dropped and other interfaces simplified.

I am inclined to remove the `AgentService` interface, but what do others think?

acoburn:
Resolves #512

This also removes the `AgentAuthorizationFilter` in the HTTP layer.


core/http/src/main/java/org/trellisldp/http/core/HttpSession.java
```diff
@@ -83,4 +92,21 @@ public IRI getAgent() {
     public Instant getCreated() {
         return created;
     }
+
+    /**
+     * Create a session from a security context.
+     * @param security the security context
+     * @return the session
+     */
+    public static Session from(final SecurityContext security) {
```
ajs6f: Maybe I'm being too picky, but since this method has the potential to create an administrative session, maybe it should be `final`?

acoburn: I thought `static` methods are _ipso facto_ `final`. Or am I mistaken there?

ajs6f: No, they can be hidden (not overridden) in subclasses. But like I said, I'm not really worried about it. On second thought, I wouldn't fool with it.

PR #520, opened 2019-09-26
Adjust NoopImpl annotations

acoburn:
I have found that, with Quarkus, the CDI injection mechanism sees the no-op service implementation beans as if they have both `@Default` and `@All` qualifiers when viewed in the context of ctor injection. E.g.:

```java
@Inject
public JenaIOService(final NamespaceService namespaceService, ...) { ... }
```

leading to ambiguous dependency resolution errors. This change expands the `@Target` types of the `@NoopImplementation` stereotype and also adds an explicit `@Priority` annotation.

### DIFF
--- a/components/app/src/test/java/org/trellisldp/app/CDIServiceBundlerTest.java
+++ b/components/app/src/test/java/org/trellisldp/app/CDIServiceBundlerTest.java
@@ -48,15 +48,8 @@ class CDIServiceBundlerTest {
                                            FileMementoService.class,
                                            JenaIOService.class,
                                            LdpConstraintService.class,
-                                           NoopAuditService.class,
-                                           NoopEventService.class,
-                                           NoopNamespaceService.class,
                                            NoopProfileCache.class,
-                                           NoopResourceService.class)
-                                       .alternatives(NoopAuditService.class,
-                                           NoopEventService.class,
-                                           NoopNamespaceService.class,
-                                           NoopResourceService.class));
+                                           TestServices.class));
 
     @Inject
     private ServiceBundler serviceBundler;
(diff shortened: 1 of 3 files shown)


acoburn:
The reason for adding the `@Priority` has to do with building this up with Quarkus in cases where there are multiple types for a given injection point (e.g. `NoopNamespaceService` and `DBNamespaceService`) and those types are used for ctor injection. I found that there were two approaches to this that work: one is adding a `@Qualifier` annotation, but that is pretty much at odds with the `@Alternative` annotation we have already. The other approach (in this PR) is to add a high-value (i.e. late-matching) `@Priority` annotation. This way, a lower value (e.g. `APPLICATION` level) annotation would match first and we can avoid the ambiguous dependency injection scenarios. FWIW, I also tried using the `@Specializes` approach, but that didn't work.

ajs6f:
I have no problem with this PR as an immediate move, but I think we should change to "no op as qualifier". I pushed the stereotype in, and it may very well be that I was premature, that we hadn't seen enough of the type system's behavior to be sure about how to do this. A qualifier is more verbose and more flexible, in the end.

acoburn:
Either way is fine with me -- the [first commit](https://github.com/trellis-ldp/trellis/pull/520/commits/cd8cf71de10b78c9e7cf04c58a7d1e28533dbb92) on this PR uses the qualifier pattern, and we can definitely go with that, if you think that would be better.

acoburn:
And if we go with the `@Qualifier` mechanism, I'd also remove the `@Alternative` and `@Stereotype` annotations.

ajs6f:
Yes, I 'd prefer that.

acoburn:
FWIW, with this change I can now make implicit use of the `BaseServiceBundler` class in `trellis-app`. Previously, I wasn't able to use that artifact with Quarkus because of these ambiguous dependency issues.

PR #523, opened 2019-09-30
Adjust WAC inheritance algorithm

### LINKED ISSUE #522: Update WebAC inheritance algorithm
The current WebAC inheritance algorithm is slightly incorrect. It should stop at the first ACL resource, not the first ACL resource with an acl:default property. See https://github.com/solid/specification/issues/55#issuecomment-536092174 for clarification.

acoburn:
Resolves #522

This changes the inheritance algorithm to be in line with the SOLID
WebAccessControl specification. In particular, the WAC inheritance will
now stop at the first ACL document it encounters, not the first ACL
document with an acl:default property.


acoburn:
That has always been the case: that only a single ACL document would ever govern the access rules for a resource. The change here relates to where to stop when looking for that ACL document: whether it should be at the first instance (the case with this PR) or at the first instance with acl:default properties (as is currently the case)

PR #528, opened 2019-10-04
Add quarkus-based project

acoburn:
This adds a quarkus/microprofile project to trellis

Related to #470

### DIFF
--- a/platform/quarkus/src/main/java/org/trellisldp/quarkus/TrellisApplication.java
+++ b/platform/quarkus/src/main/java/org/trellisldp/quarkus/TrellisApplication.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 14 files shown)


platform/quarkus/src/main/java/org/trellisldp/quarkus/TrellisApplication.java
```diff
@@ -0,0 +1,53 @@
...
+@ApplicationPath("/")
+@ApplicationScoped
+public class TrellisApplication extends Application {
+
+    private static final Logger LOGGER = getLogger(TrellisApplication.class);
+
+    @PostConstruct
+    void init() {
+        final String name = "Trellis Database Application";
+        try (final InputStream resourceStream = Thread.currentThread().getContextClassLoader()
```
jholleran: Could you use the new AppUtils.printBanner here?

acoburn: Yes, definitely. Plus, this is a triple store app, so thanks for calling my attention to this!

PR #536, opened 2019-10-07
Change ActivityStreamService to EventSerializationService

### LINKED ISSUE #531: Review ActivityStreamService interface
The `ActivityStreamService` interface is very simple, including only a single method:

```java
Optional<String> serialize(Event);
```

Yet there are two concerns I would like to raise for discussion. 

First, the name of the interface could be improved. While it is true that the implementations of this interface make use of the [ActivityStreams vocabulary](https://www.w3.org/TR/activitystreams-core/), it is not technically necessary to do so. Perhaps `NotificationService` or `ActivityService` or `EventSerializationService` would be better names.

Second, the `::serialize` method returns an `Optional<String>`. The `Optional` portion seems unnecessary. I think it would be better if the signature was simply:

```java
String serialize(Event);
```

acoburn:
Resolves #531

This also changes the return signature of `::serialize` to a simple `String`
instead of `Optional<String>`.


core/api/src/main/java/org/trellisldp/api/NoopEventSerializationService.java
```diff
@@ -13,15 +13,17 @@
...
+public class NoopEventSerializationService implements EventSerializationService {
     @Override
-    public Optional<String> serialize(final Event event) {
-        return Optional.empty();
+    public String serialize(final Event event) {
+        return "{" +
+            "\n  \"@context\": \"https://www.w3.org/ns/activitystreams\"" +
+            event.getTarget().map(obj -> ",\n  \"object\": \"" + obj.getIRIString() + "\"").orElse("") +
+            ",\n  \"id\": \"" + event.getIdentifier().getIRIString() + "\"" +
+            "\n}";
```
jholleran: Should this have a "type"?

acoburn: Excellent idea. I'll add one.

PR #553, opened 2019-10-18
Address some PMD warnings

acoburn:


auth/basic/src/main/java/org/trellisldp/auth/basic/Credentials.java
```diff
@@ -14,10 +14,15 @@
...
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.slf4j.LoggerFactory.getLogger;
 
 import java.util.Base64;
 
+import org.slf4j.Logger;
+
 public class Credentials {
+    private static Logger LOGGER = getLogger(Credentials.class);
```
acoburn: This should be `final`

components/test/src/main/java/org/trellisldp/test/AbstractResourceServiceTests.java
```diff
@@ -24,5 +24,6 @@
      * Get the ResourceService.
      * @return the resource service
      */
+    @Override
     public abstract ResourceService getResourceService();
```
jholleran: Just wondering if this abstract class is needed as it doesn't implemented any methods or have any common code.

acoburn: @jholleran that's a very good point, and the answer is probably that we don't really need that class. What we should do is probably take a broader look at the structure of the test suite. While I think the use of all these JUnit5 `@Nested` classes is very clever, it may not be the best structure for a generalizable test suite. I'll open a separate issue to review the test component.

PR #558, opened 2019-10-21
Close resources after use in test suite

acoburn:
Most of the Graph and Dataset objects in the trellis-test codebase are
not closed. This updates those tests to close the resources.

### DIFF
--- a/buildtools/src/main/resources/pmd/pmd.xml
+++ b/buildtools/src/main/resources/pmd/pmd.xml
@@ -9,7 +9,10 @@
         Trellis PMD rules
     </description>
 
-    <rule ref="category/java/bestpractices.xml"/>
+    <rule ref="category/java/bestpractices.xml">
+      <exclude name="JUnitTestContainsTooManyAsserts"/>
+      <exclude name="JUnitAssertionsShouldIncludeMessage"/>
+    </rule>
     <rule ref="category/java/multithreading.xml">
       <exclude name="UseConcurrentHashMap"/>
     </rule>
--- a/components/test/src/main/java/org/trellisldp/test/InMemoryBinaryService.java
+++ b/components/test/src/main/java/org/trellisldp/test/InMemoryBinaryService.java
@@ -81,6 +81,11 @@ public class InMemoryBinaryService implements BinaryService {
 
         private final byte[] data;
 
+        /**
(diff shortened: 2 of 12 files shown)


buildtools/src/main/resources/pmd/pmd.xml
```diff
@@ -9,7 +9,10 @@
         Trellis PMD rules
     </description>
 
-    <rule ref="category/java/bestpractices.xml"/>
+    <rule ref="category/java/bestpractices.xml">
+      <exclude name="JUnitTestContainsTooManyAsserts"/>
+      <exclude name="JUnitAssertionsShouldIncludeMessage"/>
```
ajs6f: Is there a PMD switch for warning only? It might be good for this one. Most of the assertions are fairly self-explanatory, but messages _never_ hurt.

components/test/src/main/java/org/trellisldp/test/InMemoryBinaryService.java
```diff
@@ -81,8 +81,8 @@ public String generateIdentifier() {
 
         private final byte[] data;
 
-        private InMemoryBinary(final byte[] data) {
-            this.data = data;
+        private InMemoryBinary(final InputStream data) throws IOException {
```
ajs6f: I think this should be accompanied by a Javadoc note that the stream is consumed in this ctor.

buildtools/src/main/resources/pmd/pmd.xml
acoburn: The reason I turned this one off is because the warnings were for instances where there _were_ messages. I think the PMD code (and we're using the latest release) is looking for JUnit4 constructions where messages appear in the first position, unlike in JUnit5 constructions where messages appear at the end.

components/test/src/main/java/org/trellisldp/test/InMemoryBinaryService.java
acoburn: Good idea. I'll add that

buildtools/src/main/resources/pmd/pmd.xml
ajs6f: Ahhhh, cool. Didn't know that.

PR #570, opened 2019-11-08
Update quarkus and build image

acoburn:


acoburn:
The cause/context was slightly different, but I'm employing the same technique to address it.

acoburn:
Actually, after thinking about this for a few minutes, you're right, the issue was exactly the same.

PR #572, opened 2019-11-08
Replace javax dependencies with jakarta

### LINKED ISSUE #571: Update relevant javax.* dependencies to jakarta.*
Ideally all of the javax.* dependencies will start using jakarta.*, but the priority should be for `inject` and `cdi-api`

acoburn:
Resolves #571

### DIFF
--- a/auth/basic/build.gradle
+++ b/auth/basic/build.gradle
@@ -9,15 +9,15 @@ ext {
 }
 
 dependencies {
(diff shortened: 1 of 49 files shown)


acoburn:
I'm going to let this PR sit until early next week, since once it is merged, we'll need to apply similar changes to downstream repos, and I'd rather not do that on a Friday night

PR #620, opened 2019-12-15
Ensure that LDP type triples are not duplicated

### LINKED ISSUE #619: Ensure that ldp type triples are not duplicated
When a back-end dynamically generates ldp type triples, corresponding to an interaction model (now the default), there are operations that can lead to those triples being duplicated. They should simply be filtered out on writes.

acoburn:
Resolves #619

### DIFF
--- a/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
+++ b/components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
@@ -37,7 +37,6 @@ import static org.trellisldp.triplestore.TriplestoreUtils.SUBJECT;
 import static org.trellisldp.triplestore.TriplestoreUtils.asJenaDataset;
 import static org.trellisldp.triplestore.TriplestoreUtils.getInstance;
 import static org.trellisldp.triplestore.TriplestoreUtils.getObject;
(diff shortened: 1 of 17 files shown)


components/triplestore/src/main/java/org/trellisldp/triplestore/TriplestoreResourceService.java
```diff
@@ -273,6 +274,7 @@ private UpdateRequest buildUpdateRequest(final IRI identifier, final Literal tim
             dataset.stream().filter(q -> q.getGraphName().filter(PreferServerManaged::equals).isPresent())
                     .map(rdf::asJenaQuad).forEach(sink::addQuad);
             dataset.getGraph(PreferUserManaged).ifPresent(g -> g.stream()
+                    .filter(t -> !isLdpTypeTriple(t, identifier))
```
jholleran: Is it safe to filter all the LDP types here?

acoburn: Yes because all LDP types are going to be, effectively, "server managed", so we want to keep them out of the user managed graph.

acoburn:
I believe there is a better, cleaner way to resolve this -- a way that keeps the server-managed and user-managed graphs completely separate and yet still works within the current interface structures. I'm going to work on that approach, but it might take a little longer to implement.

PR #656, opened 2020-02-01
Adjust snyk integration

acoburn:


acoburn:
The exclude flag doesn't appear to work for gradle multi-project builds.

PR #817, opened 2020-05-07
WebId-OIDC Support

rodant:
# Introduction
The purpose of this PR is to implement support for the WebId-OIDC standard used by the Solid project. This would make possible an integration between Trellis and the Node Solid Server (NSS) like: A User accesses a resource in his POD hosted on Trellis (Resource Server and Relying Party) and uses his authentication data presented by a Solid client (Presenter) after a login step against a NSS server (Identity Provider).

**I'm not an specialist in OAuth and OIDC protocols, please correct me if I miss something here.**
The solid-auth-client library sends a so called Proof of Possession (PoP) token in the `Authorization` header as `Bearer` token. 
This PoP token is constructed from my point of view according to the [WebId-OIDC Spec](https://github.com/solid/webid-oidc-spec#securing-tokens-for-multiple-resource-servers). 
In short the NSS server issues a JWT token on user login, which is embedded in a client generated token as `id_token` claim. 
The `id_token` contains additionally public key information from the client for a signature validation aligning to the JWK specification. The following is an example of a PoP token extracted from requests an example Solid App sends:

```
{"alg":"RS256"}.{ "iss":"7bc774a71455546534c821686f2a5993", "aud":"https://orisha2.solid.community", "exp":1583334972, "iat":1583331372, "id_token":<embedded-id-token>, "token_type":"pop"}.<signature>
```

The embedded `id_token` is:
`{"alg":"RS256", "kid":"m6hdkJtyAJM"}. { "iss": "https://solid.community", "sub": "https://orisha2.solid.community/profile/card#me", "aud": "7bc774a71455546534c821686f2a5993", "exp": 1574592332, "iat": 1573382732, "jti": "347676826d25d3d4", "nonce": "Ww9_k9r7cpiuL495V1X7Pnt5TvRsMJ0oQbxK5vGv-gs", "azp": "7bc774a71455546534c821686f2a5993", "cnf": { "jwk": { "alg": "RS256", "e": "AQAB", "ext": true, "key_ops": [ "verify" ], "kty": "RSA", "n": "jocytHgM-3L7P3PvUlIMXsaP9TAkQwun7HXd1QZ9eCcPiWXcu070hQx0ogNEg1uMArBGwzc8K4PJe0N8i4A2fpvLK8uHVOnSTaVXN8dPSRXlsFtLgRqBXxHsGXdjdRqKyasB4dxeYMKhajiJjr7P5YkBSyGL_BoKBBX5xC7cfZqmwbKYJ2kiKNF5EOp-NH6V9QlYWT2Yf6347s3Z5ltUPGUG30cxb3zu4gd9mV-kSO3jHhQ0u02n5yhQVrUe8jeOEGt_srPystoGZDW3gN_Isezf1YLpj7TEmLn8pbBRyNpi8iqV5lYmd2HAisCijRE5X2veTpbqgRXL1HXMZdaKBw" } }, "at_hash": "zihakOfULQlcTG-y0xXNyQ" }`

# Implementation
Besides the usual JWS checks and the WebId derivation, [WebId-OIDC](https://github.com/solid/webid-oidc-spec) requires several additional checks due to the decentralized multi-party nature of the use case. The NSS is the reference implementation for a Solid backend and does the following steps in accordance with the specification (the steps below link to the corresponding code lines in the NSS implementation):

- [WebId confirmation](https://github.com/solid/oidc-auth-manager/blob/a8b6e5706adf06790c9aecf7a1f6a4e55c28dc10/src/oidc-manager.js#L369)
  - extract WebId from claims: take the webid claim, if not present take the sub claim if it is an URL
  - [check the issuer claim](https://github.com/solid/oidc-auth-manager/blob/a8b6e5706adf06790c9aecf7a1f6a4e55c28dc10/src/oidc-manager.js#L463)  is present and that the WebId's origin is a sub domain of the issuer domain, including protocol match
- [Proof of Possession Checks](https://github.com/solid/oidc-rs/blob/4f873ea826c39655450999fcfa785757786eaa71/src/AuthenticatedRequest.js#L232)
  - [validate PoP Token](https://github.com/solid/oidc-rs/blob/4f873ea826c39655450999fcfa785757786eaa71/src/AuthenticatedRequest.js#L289)
    - check it is a PoP Token 
    - [verify PoP Token Signature and PoP Issuer matches with audience in Id-Token](https://github.com/solid/oidc-rs/blob/4f873ea826c39655450999fcfa785757786eaa71/src/PoPToken.js#L113)
    - [resolve Provider Key for verification of the Id-Token signature (OIDC Provider Metadata Discovery)](https://github.com/solid/oidc-rs/blob/4f873ea826c39655450999fcfa785757786eaa71/src/AuthenticatedRequest.js#L520)
    - [verify Id Token Signature](https://github.com/solid/oidc-rs/blob/4f873ea826c39655450999fcfa785757786eaa71/src/PoPToken.js#L240)
    - [verify Expiration of both the PoP and Id-Tokens](https://github.com/solid/oidc-rs/blob/4f873ea826c39655450999fcfa785757786eaa71/src/PoPToken.js#L83)

Our implementation reassembles these steps in the new `WebIdOIDCAuthenticator` class.

The WebId-OIDC support is configurable as the other JWT based authentication possibilities. Additionally to the `jwt` flag the new `webIdOIDC` node must be defined under the node `jwt` to activate the feature. This configuration node introduced by the PR has the form:
```
webIdOIDC:
  enabled: boolean
  cacheSize: int
  cacheExpireDays: int
```
The cache in this feature is used for the known WebId-OIDC provider keys, fetched on demand according to the [OIDC provider metadata discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig).

**References:**
* <https://openid.net/connect/>
* <https://github.com/solid/webid-oidc-spec>


acoburn:
@rodant thanks for this! The Solid WebID-OIDC protocol is undergoing quite a bit of work at the moment, so it may be premature to add this implementation now.

The direction of that work (I'm actually very much involved in it) is moving towards using [DPoP](https://tools.ietf.org/html/draft-fett-oauth-dpop-04) as a proofing mechanism rather than this Solid-specific token format with a wrapped `id_token` claim. The corresponding solid-auth-client is also in the process of moving in this direction.

Once the WebID-OIDC protocol is formalized, as part of the overall formalization of the [Solid specification](https://github.com/solid/specification), that would be a great time to add this feature.

rodant:
Hello, thanks for your really fast feedback and the background information. I can understand your decision till some point, I still must take a look to DPoP. I think we (an emerging startup in Germany) will continue our path extending and improving Trellis in general. I think it is possible to use Trellis as Solid backend in a near future and help people with concrete solutions. Best wishes.

acoburn:
Yes, DPoP is a much simpler and more secure mechanism. It's also been taken up by the OAuth2 working group, so it's on its way to becoming pretty significant, so it's worth looking into.

There are two parts to DPoP, which are based on an ephemeral keypair controlled by the app: (1) a thumbprint of the public key gets added into the access token in the `cnf` claim (the mechanism for this is by sending a DPoP header to the /token endpoint of the OIDC server). This is what the resource server (e.g. Trellis) receives in the `Authorization` header:

```javascript
{
    "sub": "https://webid.example/profile#i",
    "iss": "https://idp.example/",
    "azp": "https://app.example/",
    "cnf": {
        "jkt": "0ZcOCORZNYy-DWpqq30jZyJGHTN0d2HglBV3uiguA4I" <-- thumbprint of public key
    }
}
```

Note, that the auth header takes this form: `Authorization: DPoP <JWT-signed-token>`

(2) The second part is a special DPoP header `DPoP: <JWT-signed-token>`. That header is also a JWT, but it is created by the client application, and signed with the private key that only the client knows about (there are ways to keep that key private -- don't put it in localStorage). That JWT has this structure:

```javascript
// header
{
    "typ": "dpop+jwt",
    "alg": "ES256",
    "jwk": {  <-- embedded public key in JWK format
        "kty": "EC",
        "x": "base64-encoded-x-value",
        "y": "base64-encoded-y-value",
        "crv": "P-256"
    }
}
// body
{
    "jti": "unique-value",
    "htm": "GET", <-- HTTP method used in request
    "htu": "https://pod.example/some/resource", <-- URL of the request
    "iat": <timestamp>,
    "exp": <timestamp>
}
```

The resource server conducts a number of checks on these two headers:
1) the thumbprint in the access token must match the embedded public key in the DPoP token
2) the signature of the DPoP token must validate with the embedded public key
3) the DPoP token must not have expired
4) the method and URL in the DPoP claims must match the actual request

All of this is attempting to address the issue of token exfiltration, which is a known limitation of OIDC and JWT tokens, and which is a serious issue in a distributed context such as Solid (which is why Solid is so interested in moving in this direction)

---

Suggestions for you:

Since this PR focuses on the trellis-auth-oauth module and dropwizard, I'll assume that this is what you are especially interested in using. Here's how you can achieve that without changing any code in Trellis.

First, start with a dropwizard-based project like the one here: https://github.com/trellis-ldp/trellis/tree/master/platform/dropwizard. Just copy that code and use whatever build system you prefer, depending on the latest released Trellis artifacts.

By default, the `trellis-dropwizard` component [brings in](https://github.com/trellis-ldp/trellis/blob/master/components/dropwizard/src/main/java/org/trellisldp/dropwizard/TrellisUtils.java#L89-L105) the `trellis-auth-basic` and `trellis-auth-oauth` filters. Turn both of those off (via configuration) -- you could also leave them on, but it might be easier to just disable them.

Then, override the `AbstractTrellisApplication::initialize` [method](https://github.com/trellis-ldp/trellis/blob/master/components/dropwizard/src/main/java/org/trellisldp/dropwizard/AbstractTrellisApplication.java#L101-L103) in your custom project. By default, that's a no-op method, but you will use it to register an auth provider via:

```java
environment.jersey().register(new MySolidAuthFilter());
```

And the `MySolidAuthFilter` is yours to implement. It can be as simple or complex as you like, but it's just a JAX-RS [ContainerRequestFilter](https://docs.oracle.com/javaee/7/api/javax/ws/rs/container/ContainerRequestFilter.html). That's where you do your token parsing and validation, and where much of the code you provided here would go.

That should get you well on your way.

rodant:
Many thanks for the explanations, I appreciate it, it is great to get a summary of the DPoP specification, now I can look into the official document more easily. I see some similarities with PoP and for sure it is an advantage from the security point of view to restrict the token to just one method-resource combination, but on the other hand it also has a performance impact, the network traffic also increases considerably. You the people involved in the process have for sure considered this aspect, but it is something pops up in the mind.

For me it is not clear yet how the access and the DPoP token get combined. I'm also missing a check involving the OIDC server's keys, analogous to the id-token signature check in PoP. But I can try to find the answers in the specification.

acoburn:
> but on the other hand [DPoP] also has a performance impact, the network traffic also increases considerably

This is not entirely correct. While DPoP introduces a second HTTP header, you are exchanging one large header for two smaller headers; the overall byte size will be about the same, assuming you have the same signing algorithm. (The single Solid token wraps a second token, so the overall size of that one token is effectively the same as two separate tokens). One could argue that, under HTTP/2 with header compression, repeatedly using the same token results in less network overhead, though you still have some of that with DPoP (for the bearer token -- the DPoP token would have to be sent each time, but that's only about ~0.5 KB, uncompressed). 

If, however, you are concerned about byte size, you should be using elliptic curves and not RSA keypairs. ECDSA signatures will be a fraction of the size of RSA signatures: roughly 1/3 the size with equivalent levels of security. EC keys are also fast to create (~40 ms to create the keypair in a browser; compare that to 1-2 seconds to create RSA keypairs) and fast to compute signatures (~1 ms), so I would argue that the overall performance impact of DPoP is negligible.

The bigger issue, though is that DPoP has a specification behind it: you can find off-the-shelf software the supports it, and there are multiple implementations in the wild. The (legacy) Solid token format only exists with NSS and NSS clones, and in order to support it, you must write bespoke token parsers and rely on particular Solid-specific libraries to generate the wrapped PoP token format. Plus, once https://github.com/solid/node-solid-server/pull/1427 is merged, NSS will support DPoP tokens, too.

rodant:
My concerns about the traffic impact were different, maybe I wasn't clear enough. In PoP the client contacts the OIDC provider (OP) once at the beginning of a session with an specific Resource Server (RS), after that it has a valid PoP token for the used and can request any resource on the RS as log as the token is valid. The RS can autonomously decide according to the WAC rules if the HTTP method-resource combination is allowed. If I got it right, in a DPoP scenario, the client needs to contact the OP at least for every new combination (method, resource), this is the traffic increase I meant. Moreover the resilience is also affected, the client has stronger dependency not only on the availability and latency of the RS, but also on the OP.

These were my thoughts, but again, I still muss study the specification, maybe I'm missing a relevant aspect.

acoburn:
> in a DPoP scenario, the client needs to contact the OP at least for every new combination (method, resource)

No, a client only interacts with an OP to fetch an access token. This happens once at the beginning of a session (and subsequently to refresh expired tokens). Creation of the DPoP header itself does not involve the OP at all. In fact, the number of interactions between client and OP is the same with the Solid PoP mechanism and with DPoP, since in both cases the proofing mechanism is computed by the client, independently of the OP.

In other words, the number of network requests is the same with or without DPoP.

rodant:
Fine, thanks again for the explanation. I think I must read thoroughly the specification ;--) Good luck!

acoburn:
Closing this as won't fix, but I hope I have provided enough of a path for you to implement this in a downstream application

rodant:
> Closing this as won't fix, but I hope I have provided enough of a path for you to implement this in a downstream application

Yes, thanks again for the valuable hints and the productive exchange. I'll follow your recommendations for our application and we'll stay willing to contribute to Trellis in the future.

rodant:
> @rodant thanks for this! The Solid WebID-OIDC protocol is undergoing quite a bit of work at the moment, so it may be premature to add this implementation now.
> 
> The direction of that work (I'm actually very much involved in it) is moving towards using [DPoP](https://tools.ietf.org/html/draft-fett-oauth-dpop-04) as a proofing mechanism rather than this Solid-specific token format with a wrapped `id_token` claim. The corresponding solid-auth-client is also in the process of moving in this direction.
> 
> Once the WebID-OIDC protocol is formalized, as part of the overall formalization of the [Solid specification](https://github.com/solid/specification), that would be a great time to add this feature.

Hello Aaron, let me ask you something more. Do you know a road map for DPoP and the new WebId-OIDC?

acoburn:
@rodant I don't believe that there is a definitive roadmap I can point you to. That said, the new WebID-OIDC spec is being actively discussed in the [Solid Authentication Panel](https://github.com/solid/authentication-panel). Meeting notes are posted in that repo and can give you an idea of the direction that group is going.

PR #821, opened 2020-05-08
Allow WebAC checks to be disabled at runtime via a configuration

### LINKED ISSUE #820: Allow WebAC to be configured to disable checks at runtime
It would be very useful to be able to turn off the WAC checks at runtime using a configuration. The default of this is on.

jholleran:
Resolves #820

### DIFF
--- a/components/webac/src/main/java/org/trellisldp/webac/WebAcFilter.java
+++ b/components/webac/src/main/java/org/trellisldp/webac/WebAcFilter.java
@@ -166,7 +166,9 @@ public class WebAcFilter implements ContainerRequestFilter, ContainerResponseFil
      * @param realm the authentication realm
      * @param scope the authentication scope
      * @param baseUrl the base URL, may be null
+     * @deprecated This constructor should not be used and will be removed in a future release
      */
+    @Deprecated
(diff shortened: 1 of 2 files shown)


components/webac/src/main/java/org/trellisldp/webac/WebAcFilter.java
```diff
@@ -149,7 +154,22 @@ private WebAcFilter(final WebAcService accessService, final Config config) {
...
+    /**
+     * Create a WebAc-based auth filter.
+     *
+     * @param accessService the access service
+     * @param challengeTypes the WWW-Authenticate challenge types
+     * @param realm the authentication realm
+     * @param scope the authentication scope
+     * @param baseUrl the base URL, may be null
+     */
+    public WebAcFilter(final WebAcService accessService, final List<String> challengeTypes,
```
acoburn, line 170: The next release will be a breaking change anyway, so there's not a huge need to add this ctor (perhaps add a `@Deprecated` annotation so that we can remove it entirely later)

jholleran: Added the deprecated annotation and a comment.

PR #891, opened 2020-06-01
Use Java11 for sonar scan

acoburn:
This uses Java11 for sonar integration. It also removes coveralls


acoburn:
Even with the new sonar plugin, Java11 doesn't work so well.

PR #1022, opened 2020-08-20
Use java11 for coveralls

acoburn:


acoburn:
Travis + Java11 + Coveralls hits a TLS 1.3 error in the transport layer. We should try this inside of a GH Action workflow

